PR #32, opened 2020-04-22
use example env files for quicker setup

dekvall:
A very small thing but i think that this makes the (already very slick) setup process even easier


psikoi:
ooooh good idea, I'll check it out in a bit, when I'm on the computer, thanksss

psikoi:
Merged, thanks :)

PR #43, opened 2020-05-01
Fix typo in 'reasign'

emerald000:
Typos fix.

### DIFF
--- a/app/src/pages/Player/Player.jsx
+++ b/app/src/pages/Player/Player.jsx
@@ -47,7 +47,7 @@ const PERIOD_SELECTOR_OPTIONS = [
 
 const MENU_OPTIONS = [
   {
-    label: 'Reasign player type',
+    label: 'Reassign player type',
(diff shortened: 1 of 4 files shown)


psikoi:
Oops! Thanks for noticing, could you squash those into a single commit? And with a more appropriate commit message, like "Fix reassign typo"

emerald000:
Can you just squash and merge? I was hoping to just use the Github edit tools and not having to download everything with git just for this.

psikoi:
Sure, are you sure these are the only mentions of that word? I'm not currently on the computer to check

emerald000:
Yeah, those are the only 3 instances in the repo.

emerald000:
Turns out Github sucks and didn't give all the instances.

So I ended up doing things correctly anyway.

psikoi:
I was about to double check myself, seems correct now! Thanks

PR #63, opened 2020-05-03
Ctrlclick buttons

ClumsyPenguin:


psikoi:
We can yes

ClumsyPenguin:
quick verify using only anchor tag will translate it to this?
```html
<a href="https://github.com/psikoi/wise-old-man" className="-github-btn" role="button">
   <img src="/img/icons/github.svg" alt="github icon" /> 
    Contribute on Github
</a>
```

psikoi:
There's a lot of buttons being used where it should be anchors, so it would be best to go through every usage of <Button> and see if its well used

TimvdLippe:
> it may be wiser to make a component for this, so we can easily implement this behaviour at more places?

I recommend against this. You will likely run into the problem where the abstraction you create is more limiting than helping. Given that they are relatively simple navigation elements, a regular `<a>` should suffice.

ClumsyPenguin:
i fail to properly allign the text with the icons, if i change the icon padding-top it makes the icon smaller ?! weird... if someone has a suggestion on how to allign this. Shoot!

TimvdLippe:
I think you can use flexbox and `align-items: center` for that: https://stackoverflow.com/a/489394

ClumsyPenguin:
> 
> 
> I think you can use flexbox and `align-items: center` for that: https://stackoverflow.com/a/489394

i thought because it was allready being used by the parent that the child (icon and text) inherent this properties, but this what not the case. thanks for the suggestion

psikoi:
Thanks!

Issue #64, opened 2020-05-03
NumberLabel - Mobile support


psikoi:
Currently the NumberLabel component (that displays formatted values like +13.7m and +2345k) has a hover effect that shows the full number, but this of course cannot be done on mobile so it would be nice to have a popup on tap, that shows the full number for mobile devices

PR #70, opened 2020-05-04
Add popup tot NumberLabel so mobile users can see full value #64

RoelHermans:


psikoi:
You should create a branch for new code, and then pull request that branch

PR #71, opened 2020-05-04
Add popup to NumberLabel so mobile users can see full value #64

### LINKED ISSUE #64: NumberLabel - Mobile support

RoelHermans:
Fixes #64

### DIFF
--- a/app/src/components/NumberLabel/NumberLabel.jsx
+++ b/app/src/components/NumberLabel/NumberLabel.jsx
@@ -33,7 +33,7 @@ function NumberLabel({ value, isColored, isSigned, lowThreshold }) {
 
   const finalValue = isSigned ? signNumber(value, formattedValue) : formattedValue;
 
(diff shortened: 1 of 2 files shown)


psikoi:
There's something wrong with the issues, it's showing you're submmiting 4 when you're really only submitting 1.

Can you try and rebase to fix it? Join the discord if you need any help

psikoi:
Added a commit to rename the css class of that component (it was outdated, NumberLabel used to be called SpecialNumber) and chose a lighter background color for the popup to make it stand out more.

Also rebased and force pushed your code. It should now be ready to merge.

psikoi:
Thanks :)

PR #75, opened 2020-05-04
Show player type on icon hover

Knutakir:
This is for showing the player type on hovering the icon in a specific player page.

### DIFF
--- a/app/src/utils/player.js
+++ b/app/src/utils/player.js
@@ -23,3 +23,13 @@ export function getOfficialHiscoresUrl(player) {
 
   return `https://secure.runescape.com/m=hiscore_oldschool${suffix}/hiscorepersonal.ws?user1=${username}`;
 }
+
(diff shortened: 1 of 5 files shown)


app/src/utils/player.js
```diff
@@ -23,3 +23,13 @@ export function getOfficialHiscoresUrl(player) {
 
   return `https://secure.runescape.com/m=hiscore_oldschool${suffix}/hiscorepersonal.ws?user1=${username}`;
 }
+
+export function getTooltip(type) {
```
psikoi, line 27: Could you rename this to `getPlayerTooltip` please? As it's referenced as "getTooltip" anywhere on the app, it's a bit ambiguous.

Knutakir: Will do

psikoi:
Looks good!

PR #77, opened 2020-05-05
Add virtual levels

Knutakir:
It is implemented as `99 (126)` (as suggested by another member in the Discord server).
It does show the virtual level only if the level is above 99. Also added a virtual total level that is present if the user has any virtual levels.


psikoi:
Reading on mobile but that looks perfect :D

psikoi:
Added the option to show virtual/regular levels based on a dropdown selection (instead of always showing virtuals too), as I didn't like the way the table looked with both at the same time.

psikoi:
Thanks!

PR #84, opened 2020-05-06
Update Unknown Achievements

bolinkd:
## Description
Added a new section on the server the check for old achievements when importing from cml or updating the player

## Notes
The biggest issue I can see here is that the date we display will be the first time we see a snapshot after a user got 99
So even if they got it on one day, it may appear a couple weeks later without any real way to check, the only thing we know for sure is that it can not be before the date we have posted

Another note: A user that updates may appear to have their 99 on the date of the first update, this will be fixed after CML imports and the old snapshots are checked for dates.

## Screenshots
Screen Shot 2020-05-05 at 20 11 43


.gitignore
```diff
@@ -23,6 +23,7 @@ chrome-user-data
 .vscode
 *.swp
 *.swo
+*.code-workspace
```
bolinkd, line 26: make sure my vscode workspace isnt added :)

server/src/database/migrations/20200503201112-add-lastUpdatedAchievementsAt-player.js
```diff
@@ -0,0 +1,23 @@
...
+      Promise.all([
+        queryInterface.addColumn(
+          'players',
+          'lastUpdatedAchievementsAt',
+          {
+            type: Sequelize.DATE
+          },
+          { transaction }
+        ),
+        queryInterface.bulkDelete('achievements', {}, { transaction })
```
bolinkd, line 13: Deleting achievements and starting over seemed like a better way to go, will have to see how this meshes with your new achievements code

bolinkd: I could remove this and do a check specifically for Date(0), but its hard to check if a date is before the previous one if its as far back as possible

server/src/api/modules/achievements/achievement.service.js
```diff
@@ -1,24 +1,44 @@
...
-  ACHIEVEMENTS.forEach(a => {
-    if (a.name.includes('{skill}')) {
-      SKILLS.filter(s => s !== 'overall').forEach(skill => {
-        if (a.validate(snapshot[`${skill}Experience`])) {
-          achievements.push(a.name.replace('{skill}', skill));
+function getAchievements(snapshots, prevAchievements) {
+  const achievements = prevAchievements || {};
+
+  // Return true if d1 is after d2
+  const dateCmp = (d1, d2) => d1.getTime() > d2.getTime();
```
bolinkd, line 16: could add the date check here for Date(0) and replace if so

bolinkd:
Going to rebase this off of the changes coming to boss kc and reopen later

PR #85, opened 2020-05-06
Back-end for bulk valid username checking

Jakesterwars:
This is the first step for bulk valid username checking. The front end will need to use the returned list of names to highlight the invalid ones in the list.


app/src/redux/modules/groups/actions/create.js
```diff
@@ -40,6 +40,8 @@ export default function createGroup({ name, members }) {
     return axios
       .post(url, body)
       .then(result => dispatch(createGroupSuccess(result.data)))
-      .catch(error => dispatch(createGroupFailure(error)));
+      .catch(error => {
```
psikoi, line 43: Is there a reason why you added those brackets? They are not needed

app/src/components/Table/Table.scss
```diff
@@ -83,6 +83,10 @@
       color: white;
       font-weight: 700;
     }
+
+    &.-error {
+        color: red;
```
psikoi, line 88: Please use $negative-red for color scheme consistency.

app/src/components/TableList/TableList.scss
```diff
@@ -47,6 +47,10 @@
       color: $negative-red;
     }
 
+    &.-error {
+      color: red;
```
psikoi, line 51: Please use $negative-red for color scheme consistency.

server/src/api/modules/competitions/competition.service.js
```diff
@@ -272,12 +272,16 @@ async function create(title, metric, startsAt, endsAt, groupId, participants) {
 
   // Check if every username in the list is valid
   if (participants && participants.length > 0) {
-    for (let i = 0; i < participants.length; i += 1) {
-      if (!playerService.isValidUsername(participants[i])) {
-        throw new BadRequestError(`Invalid player username: ${participants[i]}`);
-      }
+    const usernamesCheck = await playerService.isValidUsernames(participants);
```
psikoi, line 275: Maybe this variable could be renamed to "invalidUsernames"?

**Also: isValidUsernames is not an async function so it does not need the "await" before calling it.**

server/src/api/modules/competitions/competition.service.js
```diff
@@ -364,10 +368,13 @@ async function edit(id, title, metric, startsAt, endsAt, participants, verificat
 
   if (participants) {
     // Check if every username in the list is valid
-    for (let i = 0; i < participants.length; i += 1) {
-      if (!playerService.isValidUsername(participants[i])) {
-        throw new BadRequestError(`Invalid player username: ${participants[i]}`);
-      }
+    const usernamesCheck = await playerService.isValidUsernames(participants);
```
psikoi, line 371: Maybe this variable could be renamed to "invalidUsernames"?

**Also: isValidUsernames is not an async function so it does not need the "await" before calling it.**

server/src/api/modules/groups/group.service.js
```diff
@@ -194,10 +194,13 @@ async function create(name, members) {
 
   // Check if every username in the list is valid
   if (members && members.length > 0) {
-    for (let i = 0; i < members.length; i += 1) {
-      if (!playerService.isValidUsername(members[i].username)) {
-        throw new BadRequestError(`Invalid player username: ${members[i].username}`);
-      }
+    const usernamesCheck = await playerService.isValidUsernames(members);
```
psikoi, line 197: Maybe this variable could be renamed to "invalidUsernames"?

**Also: isValidUsernames is not an async function so it does not need the "await" before calling it.**

server/src/api/modules/groups/group.service.js
```diff
@@ -255,10 +258,13 @@ async function edit(id, name, verificationCode, members) {
 
   if (members) {
     // Check if every username in the list is valid
-    for (let i = 0; i < members.length; i += 1) {
-      if (!playerService.isValidUsername(members[i].username)) {
-        throw new BadRequestError(`Invalid player username: ${members[i].username}`);
-      }
+    const usernamesCheck = await playerService.isValidUsernames(members);
```
psikoi, line 261: Maybe this variable could be renamed to "invalidUsernames"?

**Also: isValidUsernames is not an async function so it does not need the "await" before calling it.**

server/src/api/modules/players/player.service.js
```diff
@@ -27,22 +27,26 @@ function formatUsername(username) {
     .join(' ');
 }
 
-function isValidUsername(username) {
-  const formattedUsername = formatUsername(username);
+function isValidUsernames(members) {
```
psikoi, line 30: Try not to use the term member for this as "member" is a term used for groups, and since you're using this method for competitions, which uses "participant", it's incorrect. Maybe "usernames" is a more generic variable name.

**Also, is there a reason why you're changing this method to accept multiple values, instead of doing something like:** 
`
const invalidUsernames = usernames.filter(playerService.isValidUsername);
`

This way, the isValidUsername function could be used for single usernames in the future,

app/src/redux/modules/groups/actions/create.js
Jakesterwars: I had a console log in there to spit out the error for a debugging thing. I forgot to remove it when I removed the log.

app/src/components/Table/Table.scss
Jakesterwars: Will remove when I can.

app/src/components/TableList/TableList.scss
Jakesterwars: Will remove when I can.

server/src/api/modules/competitions/competition.service.js
Jakesterwars: I must've been late night reading because I thought I read async on the player service function. :) I'll remove the await.

server/src/api/modules/competitions/competition.service.js
Jakesterwars: I'll rename it.

server/src/api/modules/groups/group.service.js
Jakesterwars: Same as above.

server/src/api/modules/players/player.service.js
Jakesterwars: Posting what I said in Discord for historical sake:

> The reason I changed it that way is that there’s no reason you should need to call that function X number of times for a validation when you can send all the data once, loop, and return whatever is bad back.
You can still do a single check by passing in a single name in an array if you want still.

Jakesterwars:
Merged in with #110

PR #86, opened 2020-05-06
added math.max

ClumsyPenguin:


psikoi:
Hum this should be handled on the api actually

psikoi:
Fixed (server-side) in https://github.com/psikoi/wise-old-man/commit/8f72bdfefce257bfa6bae42f4d2d2dd143371c3e

Thank you tho!

PR #73, opened 2020-05-04
Track Boss Kc and Activity Score

psikoi:
# Boss/Activity Tracking

This PR adds boss and activity tracking to the API and the App, with this, also comes boss/activity competitions, achievements, records, etc.

**API To-do:**
- [x] Add boss and activity fields in snapshot model
- [x] Fetch boss and activity data from RS hiscores
- [x] Save fetched boss and activity data into snapshots
- [x] Add bosses and activities to snapshot service
- [x] Add bosses and activities to achievement service
- [x] Add 500,1000,5000,10000 boss killcount achievements
- [x] Add 100 kc in every boss achievement
- [x] Add 500,1000,5000,10000 activity score achievements
- [x] Add bosses and activities to deltas
- [x] Add bosses and activities to records
- [x] Improve record create/update efficiency
- [x] Add bosses and activities to competitions
- [x] Update API documentarion

**APP To-do:**
- [x] Selector Component: Add search functionality
- [x] Home Page: Add new feature cards
- [x] Current Top Page: Add selector (Skilling / PvM / Activities)
- [x] Records Page: Add selector (Skilling / PvM / Activities)
- [x] Competitions Page: Add selector (Skilling / PvM / Activities)
- [x] Create Competition Page: Add new metrics
- [x] Edit Competition Page: Add new metrics
- [x] Competition Page: Add new widget images
- [x] Competition Page: Change Info panel label dynamically
- [x] Competition Page: Change Top Player widget
- [x] Competition Page: Change Total Gained widget
- [x] Player Page: Overview Add selector (Skilling / PvM / Activities)
- [x] Player Page: Gained Add selector (Skilling / PvM / Activities)
- [x] Player Page: Gained Select metric by clicking table
- [x] Player Page: Records Add selector (Skilling / PvM / Activities)

Fixes #76

PR #88, opened 2020-05-06
Fix Runecrafting Icon on Achievements page

dabolink:
Moved runecrafting above crafting in skills list as "runecrafting" includes "crafting"

### DIFF
--- a/app/src/config/skills.js
+++ b/app/src/config/skills.js
@@ -12,6 +12,7 @@ export const SKILLS = [
   'fletching',
   'fishing',
   'firemaking',
+  'runecrafting',
   'crafting',
   'smithing',
   'mining',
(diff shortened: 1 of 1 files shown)


psikoi:
This bug was fixed in #73 as achievement skill names are now capitalized (Ex: 99 Runecrafting), so it will now fail the `'99 Runecrafting'.includes('Crafting')` validation.[ See here](https://github.com/psikoi/wise-old-man/blob/2f8ced656fe48d55235a45cf9fda6d962e006d65/app/src/pages/Player/components/PlayerAchievementsWidget/PlayerAchievementsWidget.jsx#L10)

This PR is no longer necessary, but thank you for helping out!

PR #98, opened 2020-05-07
Show combat level in the player page

Knutakir:
Show the players combat level in the player page.
It shows `-` if at least one of the levels used for calculating the combat level returns `0`. Otherwise in the range of `3` to `126`.

### Example image
image


psikoi:
It seems you're not using the right prettier/eslint configs.

Please do:

(in your project root)
`npm i`

This should install all the linting dependencies

Then you can just `ctrl + S` to re-save the formatted code.

psikoi:
Otherwise it looks good to go (if formatted)

psikoi:
Actually as @dekvall suggested on discord, this might be useful information to have on the API. Could you make the /players/:id api return `combatLevel: X`? And then use that field on the page just like you're doing

Knutakir:
👍 Will move it when I have time

psikoi:
Thanks! Sorry for the switcharoo

Jakesterwars:
Mentioned in Discord:
`As for the combat level thing, we could also make it a little cleaner than the || hell. Just set the values in an array in the function and indexOf the 0. If it shows up, return the -.
Not functionally different or performant enhancing, just looks cleaner.`

I.E.

`  if [attackLevel, strengthLevel, defenceLevel, hitpointsLevel, rangedLevel, prayerLevel, magicLevel].some(level => level === 0)) {
    return '-';
  }`

server/src/api/modules/players/player.service.js
```diff
@@ -113,8 +139,33 @@ async function getDataById(id) {
...
+  const magicLevel = getLevel(magicExperience);
+  const combatLevel = getCombatLevel(
+    attackLevel,
+    strengthLevel,
+    defenceLevel,
+    hitpointsLevel,
+    rangedLevel,
+    prayerLevel,
+    magicLevel
+  );
```
dekvall, line 166: This should probably be put in a function rather than duplicating logic

server/src/api/modules/players/player.service.js
```diff
@@ -94,8 +95,33 @@ async function getData(username) {
...
```
dekvall, line 122:
```suggestion
  const levels = [attackExperience, strengthExperience, defenceExperience, hitpointsExperience, rangedExperience, prayerExperience, magicExperience].map(xp => getLevel(xp))
  const combatLevel = getCombatLevel(...levels);
```

server/src/api/util/level.js
```diff
@@ -0,0 +1,52 @@
...
+      strengthLevel,
+      defenceLevel,
+      hitpointsLevel,
+      rangedLevel,
+      prayerLevel,
+      magicLevel
+    ].some(level => level === 0)
+  ) {
+    return '-';
+  }
```
dekvall, line 38: imo this should just be
```suggestion
  // eslint-disable-next-line prefer-rest-params
  if ([...arguments].some(l => l === 0)) {
    return '-';
  }
```

server/src/api/modules/players/player.service.js
Knutakir: That looks way better 😄

server/src/api/util/level.js
Knutakir: Yep, agree, will do this!

server/src/api/modules/players/player.service.js
Knutakir: 👍 Will look into making it less duplicating

Knutakir:
Still thinks the `getCombatLevel` function needs more refactoring.

server/src/api/util/level.js
```diff
@@ -0,0 +1,69 @@
...
+  for (index = 0; index < LEVEL_EXP.length; index += 1) {
+    if (LEVEL_EXP[index + 1] > experience) {
+      break;
+    }
+  }
+
+  return Math.min(index, 99);
+}
+
+function getCombatLevel(playerExperiences) {
```
dekvall, line 20: This still feels uncomfortably messy, maybe we can use [pick](https://underscorejs.org/#pick) to extract the fields we need and [transform](https://lodash.com/docs/4.17.15#transform) to transform the objects. Then we can extract the values when calculating. So something like
```
const combatExperience = _.pick(playerExperiences, attackExperience, ..., magicExperience);
const levels = _.transform(combatExperience, (r, v, k) => {
    r[k.replace("Experience", "Level")] = getLevel(v);
});
...
```

server/src/api/constants/levels.js
```diff
@@ -0,0 +1,131 @@
+const LEVEL_EXP = [
```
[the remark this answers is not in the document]
Knutakir: Yes, will do!

server/src/api/util/level.js
```diff
@@ -0,0 +1,72 @@
...
+  });
+
+  // Ensure all skills are set
+  if (Object.keys(levels).length !== 7) {
+    return 0;
+  }
+
+  // If the player has at least one of the stats as level 0 the calculation becomes incorrect
+  // This is due to the player not being on the Hiscores
+  if (Object.keys(levels).some(level => level === 0)) {
```
dekvall, line 56: This should be `Object.values()`

Knutakir: 👍 Yep, my bad

server/src/api/util/level.js
```diff
@@ -0,0 +1,72 @@
...
+    'magicExperience'
+  ]);
+
+  const levels = _.transform(combatExperiences, (r, v, k) => {
+    // eslint-disable-next-line no-param-reassign
+    r[k.replace('Experience', '')] = getLevel(v);
+  });
+
+  // Ensure all skills are set
+  if (Object.keys(levels).length !== 7) {
```
dekvall, line 50: I don't think this will ever be true so we can probably skip this check.

Knutakir: Probably. When I was testing I had an empty object as the `snapshot` to the player. It made the function return `NaN`. But if's unnecessary in reality I can remove it.

docs/src/configs/docs/players/entities.js
```diff
@@ -6,38 +6,43 @@ export default [
...
       },
       {
         field: 'updatedAt',
         type: 'date',
-        description: "The player's last update date.",
+        description: "The player's last update date."
       },
-    ],
+      {
+        field: 'combatLevel',
```
psikoi, line 37: Hum the combatLevel isn't really a column in the players table, it's just a calculated field in the player details endpoint

Knutakir: I must have been quick and thought it was the response fields of the `.../api/players/?id={x}` endpoint. Will remove this part.

PR #128, opened 2020-05-15
Name capitalization check

### LINKED ISSUE #45: Capital letters are not respected in display names
I started tracking as "TheLotto" but now the site says my name is Thelotto.

### LINKED ISSUE #123: Automatically correct username capitalization
Extension of #45 

Since we are currently formatting all usernames "To Be Like This", a lot of people with custom usernames aren't happy.

A suggestion that has been discussed before on discord is to use the hiscores leaderboard pages to check the correct capitlization for any given name. (Because names are correctly capitalized in those pages for some reason).

The only way that we're currently aware of to do this, is to fetch the contents of this page:

`https://secure.runescape.com/m=hiscore_oldschool/overall?table=0&user=lynx%A0titan`

Find the correct table row by checking a name match:
`lowercase(tableUsername) === lowercase(databaseUsername)`

Or by checking the row's color (it seems to be red)

The question now is:

Would this be another endpoint in the players module, or an automatic check onPlayerUpdate?(maybe with a time limit, like once every 24h)

PS: With this, some other endpoints on the API have to be changed to support unstandardized usernames.

psikoi:
Fixes #45 
Resolves #123 

- Add displayName column to player model
- Store all usernames in lower case
- Use new "standardize" method for username checking
- Fix repeated usernames in competitions/groups
- Add new capitalization assertion endpoint `POST /players/assert-name`
- Run username assertion on player registration
- Update docs for new name assertion endpoint

### DIFF
--- a/app/src/components/MembersSelector/MembersSelector.jsx
+++ b/app/src/components/MembersSelector/MembersSelector.jsx
@@ -15,9 +15,10 @@ function getTableConfig(invalidUsernames, onRemove, onSwitchRole) {
     uniqueKeySelector: row => row.username,
     columns: [
       {
-        key: 'username',
+        key: 'displayName',
+        label: 'Name',
         width: 170,
-        className: val => (isInvalid(val) ? '-negative' : '-primary')
+        className: (val, row) => (isInvalid(row ? row.username : '') ? '-negative' : '-primary')
       },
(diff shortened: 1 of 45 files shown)


psikoi:
Awaiting #134

PR #130, opened 2020-05-16
add ehp rates dep

dekvall:
Here are some rates.

### DIFF
--- a/server/package.json
+++ b/server/package.json
@@ -28,7 +28,8 @@
     "moment": "^2.24.0",
     "pg": "^7.18.2",
     "pg-hstore": "^2.3.3",
-    "sequelize": "^5.21.4"
+    "sequelize": "^5.21.4",
+    "wom-rates": "github:wise-old-man/ehp-rates#master"
   },
   "devDependencies": {
     "faker": "^4.1.0",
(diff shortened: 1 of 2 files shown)


server/package.json
```diff
@@ -28,7 +28,8 @@
     "moment": "^2.24.0",
     "pg": "^7.18.2",
     "pg-hstore": "^2.3.3",
-    "sequelize": "^5.21.4"
+    "sequelize": "^5.21.4",
+    "wom-rates": "github:wise-old-man/ehp-rates#master"
```
Knutakir, line 32: Shouldn't we import this as e.g.: `wom-ehp-rates` for less ambiguity? Since just `wom-rates` can be different things.

psikoi: We've since transfered that repo to the wise-old-man organization, and renamed the repo to ehp-rates

so it's now 

`wise-old-man/ehp-rates`

Knutakir: I mean on the left side. I agree with the repo and org name.

Knutakir: Or will it be just `ehp-rates` then?

psikoi: I think it's independant enough, so that if you wanted to use those rates in another project, it would make sense to have an ambiguous name, but i get ur point

psikoi:
Closing until Mr. @dekvall is back from hiatus.

PR #134, opened 2020-05-17
Competition API tests & enhancements

DennisSnijder:
# Context
A recent bug showed the need for tests on the competition endpoints.

# What has been done
- Competition creation API tests
- Moved the model validation to the competition service to remain consistent.
- Changed some Exceptions being thrown to return proper status codes.

### DIFF
--- a/docs/src/configs/docs/competitions/endpoints.js
+++ b/docs/src/configs/docs/competitions/endpoints.js
@@ -236,7 +236,7 @@ export default [
       },
       {
         description: 'If startsAt OR endsAt is a past date.',
-        body: { message: 'Invalid dates: All start and end dates must be in the future.' }
+        body: { message: 'Start date must be before the end date.' }
       },
       {
         description: "If one of the participant' usernames is invalid",
--- a/server/__tests__/integration/competition.test.js
+++ b/server/__tests__/integration/competition.test.js
@@ -1,24 +1,27 @@
 const supertest = require('supertest');
 const api = require('../../src/api');
-const { Player, Competition } = require('../../src/database');
+const { Player, Competition, Group, Membership } = require('../../src/database');
 const { resetDatabase } = require('../utils');
 
 const TEST_DATA = {};
 const TEST_ID = 300000;
+const TEST_GROUP_ID = 9001;
+
 const TEST_VERIFICATION_CODE = '237-221-631';
+const TEST_GROUP_VERIFICATION_CODE = '021-321-025';
 
 const request = supertest(api);
 
 beforeAll(async done => {
   await resetDatabase();
 
-  await Player.create({
+  const player1 = await Player.create({
     id: 1000000,
     username: 'test player',
-    displayName: 'Test Player'
+    displayName: 'Test Player',
   });
 
-  await Player.create({
+  const player2 = await Player.create({
     id: 200000,
     username: 'alt player',
     displayName: 'Alt Player'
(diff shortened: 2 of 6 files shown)


server/src/api/modules/competitions/competition.service.js
```diff
@@ -307,15 +309,23 @@ async function create(title, metric, startsAt, endsAt, groupId, groupVerificatio
...
+      title: sanitizedTitle,
+      metric: metric.toLowerCase(),
+      verificationCode,
+      verificationHash,
+      startsAt,
+      endsAt,
+      groupId
+    });
+  } catch (e) {
+    if(e instanceof ValidationError) {
```
DennisSnijder, line 323: @psikoi perhaps we should move the model validation to the service 🤔

DennisSnijder:
@psikoi PR is ready! 😄 

image

server/src/api/modules/competitions/competition.service.js
```diff
@@ -13,7 +13,7 @@ const {
 } = require('../../../database');
 const { durationBetween, isValidDate, isPast } = require('../../util/dates');
 const { generateVerification, verifyCode } = require('../../util/verification');
-const { BadRequestError } = require('../../errors');
+const { BadRequestError, NotFoundError } = require('../../errors');
```
psikoi, line 16: NotFound is 404, is this applicable to an id not being found?

DennisSnijder:
@psikoi added some additional tests for adding participants, removing them and creating competitions as a group. Only thing I'm not 100% covering is all of the verificationCodes in each endpoint, which feels kinda repetetive since we're using the same logic for verifying that on pretty much all calls.

tldr; Ready for review once again

image

docs/src/configs/docs/competitions/endpoints.js
```diff
@@ -236,7 +236,7 @@ export default [
       },
       {
         description: 'If startsAt OR endsAt is a past date.',
-        body: { message: 'Invalid dates: All start and end dates must be in the future.' }
+        body: { message: 'Start date must be before the end date.' }
```
psikoi, line 239: Is the previous message not relevant anymore? This doesn't seem to be just a grammar correction.

server/__tests__/integration/competition.test.js
```diff
@@ -0,0 +1,512 @@
...
+      const response = await request.get(`/api/competitions`).send();
+
+      expect(response.status).toBe(200);
+      expect(response.body.length).toBe(3);
+
+      done();
+    });
+
+    test('Search competitions ( invalid status )', async done => {
+      const response = await request.get(`/api/competitions?status=invalid-status`).send();
```
psikoi, line 253: It would be better to not hardcode the query params, check the Player search tests to see how you can do it differently ;)

server/__tests__/integration/competition.test.js
```diff
@@ -0,0 +1,512 @@
...
+      done();
+    });
+
+    test('Search competitions ( with metric )', async done => {
+      const response = await request.get(`/api/competitions?metric=OvErAlL`).send();
+
+      expect(response.status).toBe(200);
+      expect(response.body.length).toBe(3);
+
+      await response.body.map(c => expect(c.metric).toBe('overall'));
```
psikoi, line 276: why is there an `await` here?

server/__tests__/integration/competition.test.js
```diff
@@ -0,0 +1,512 @@
...
+        verificationCode: TEST_DATA.minimal.verificationCode
+      });
+
+      expect(response.status).toBe(404);
+      expect(response.body.message).toBe('Competition of id 1234 was not found.');
+
+      done();
+    });
+
+    test('Do not remove participant ( No participants givens )', async done => {
```
psikoi, line 435: "givens" 😆

server/__tests__/integration/competition.test.js
```diff
@@ -0,0 +1,512 @@
...
+
+      expect(response.status).toBe(200);
+      expect(response.body.message).toBe(
+        `Successfully removed 1 participants from competition of id: ${TEST_DATA.minimal.id}.`
+      );
+
+      done();
+    });
+
+    test('Do not remove participant ( Not in participant list )', async done => {
```
psikoi, line 473: Isn't this test repeated? line 447

docs/src/configs/docs/competitions/endpoints.js
DennisSnijder: Yup, the invalid dates prefix was caused by the model validation, since most errors don't have such error prefix. I thought I'd remove it.

server/__tests__/integration/competition.test.js
DennisSnijder: yup! seems like I was copy pasting a bit too much haha

DennisSnijder:
@psikoi remarks fixed!

docs/src/configs/docs/competitions/endpoints.js
psikoi: No I mean the original meaning of the message, all start and end dates cannot be past dates?

DennisSnijder: Yeah, sounds kinda weird, what'd you suggest? 🤔

PR #138, opened 2020-05-19
Added confirmation modal when user tries to submit empty competition/group

breeku:
Addresses https://github.com/psikoi/wise-old-man/issues/113

This is my first contribution to open source project so there's high probability that i'm doing something wrong, let me know! :)


app/src/modals/EmptyCompetitionModal/EmptyCompetitionModal.jsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Button from '../../components/Button';
+import './EmptyCompetitionModal.scss';
+
+function EmptyCompetitionModal({ onConfirm, hideModal }) {
```
psikoi, line 6: This hideModal action should be on the page that mounts this modal, like the others modals (check the ImportalPlayersModal) for reference.

The page should have the responsibility to control (hide/show) it's children. Makes the flow easier, this way components/modal can just exist, without having to worry about showing or not.

app/src/pages/CreateCompetition/CreateCompetition.jsx
```diff
@@ -224,7 +228,15 @@ function CreateCompetition() {
           )}
         </div>
         <div className="form-row form-actions">
-          <Button text="Confirm" onClick={onSubmit} loading={isSubmitting} />
+          <Button
+            text="Confirm"
+            onClick={
+              (!groupCompetition && participants.length === 0) || (groupCompetition && !selectedGroup)
```
psikoi, line 234: This could look cleaner if you declared a "isEmpty" variable before the return statement, and then used it in this validation.

`isEmpty ? showEmptyCompetitionModal : onSubmit`

app/src/modals/EmptyCompetitionModal/EmptyCompetitionModal.scss
```diff
@@ -0,0 +1,67 @@
...
+    background: $gray-00;
+    width: 300px;
+    max-width: 90%;
+    min-height: 100px;
+    box-shadow: 0px 5px 10px rgba(0, 0, 0, 0.5);
+    padding: 25px;
+    display: flex;
+    flex-direction: column;
+
+    .modal-title {
```
psikoi, line 29: Could you reduce the font size a bit? That h1 default size seems too big

app/src/modals/EmptyCompetitionModal/EmptyCompetitionModal.jsx
breeku: Not quite sure if I understood correctly, but I tried to mimic the way ImportPlayersModal is done.

psikoi: ImportPlayersModal doesnt have a hide callback:

`function ImportPlayersModal({ onConfirm, onClose })`

The hiding action part is handled by the CompetitionPage, whenever onConfirm or onCancel get called

breeku:
Thanks for welcoming! I've no idea why package.lock has changed since I only did npm install's. How should I go about solving that?

psikoi:
> Thanks for welcoming! I've no idea why package.lock has changed since I only did npm install's. How should I go about solving that?

It's probably alright, there doesn't seem to be any package versions changes, so don't worry about it. I was just wondering if you had done anything besides installs

app/src/modals/EmptyCompetitionModal/EmptyCompetitionModal.jsx
breeku: I think they both did hiding modal the same way, I had just named them different? I changed the naming scheme.
Or are you talking about handleConfirm()?

app/src/modals/EmptyCompetitionModal/EmptyCompetitionModal.scss
breeku: 2em -> 1.75em

psikoi:
> Could you make this modal also support empty member groups?

Are you willing to do this, in this PR?

breeku:
> 
> 
> > Could you make this modal also support empty member groups?
> 
> Are you willing to do this, in this PR? The current competition modal is good to merge, but I'd like the same done for groups

https://github.com/psikoi/wise-old-man/pull/138/commits/03e9920f0c8063a81ba7ee836f3236f58d9f4d9b Checks if the selected group is empty, or are you talking about something else?

psikoi:
> [03e9920](https://github.com/psikoi/wise-old-man/commit/03e9920f0c8063a81ba7ee836f3236f58d9f4d9b) Checks if the selected group is empty, or are you talking about something else?

I was talking about also adding a modal for creating groups with 0 members. As described in the original issue. Perhaps this modal could be abstracted to serve both use cases? Like ImportPlayersModal was recently

breeku:
> 
> 
> > [03e9920](https://github.com/psikoi/wise-old-man/commit/03e9920f0c8063a81ba7ee836f3236f58d9f4d9b) Checks if the selected group is empty, or are you talking about something else?
> 
> I was talking about also adding a modal for creating groups with 0 members. As described in the original issue. Perhaps this modal could be abstracted to serve both use cases? Like ImportPlayersModal was recently

Ah I see, I actually haven't looked at that at all, overlooked the issue. Abstraction sound's good though

psikoi:
Looks good! will quickly checkout locally and do a finaly test ;)

PR #139, opened 2020-05-19
Sortable table

breeku:
Addresses https://github.com/psikoi/wise-old-man/issues/136

Sorting modes are default, ascending and descending. Clicking loops through them

- [x] Functionality

- [x] CSS

### DIFF
--- a/app/src/components/Table/Table.jsx
+++ b/app/src/components/Table/Table.jsx
@@ -1,102 +1,68 @@
-import React, { useState, useEffect } from 'react';
+import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import { capitalize } from '../../utils';
 import './Table.scss';
 
-const SORT = {
-  DEFAULT: 'default',
-  ASCENDING: 'ascending',
-  DESCENDING: 'descending'
-};
-
-const DEFAULT_SORTING = { type: SORT.DEFAULT, by: '' };
-
-function getValue(row, key, get, transform) {
+function getRowValue(row, key, get, transform) {
   const value = get ? get(row) : row[key];
   return [transform ? transform(value, row) : value, value];
 }
 
-function Table({ rows, columns, highlightedIndex, onRowClicked, clickable }) {
-  const [sorting, setSorting] = useState(DEFAULT_SORTING);
+function Table({ rows, columns, uniqueKeySelector, highlightedIndex, onRowClicked }) {
+  const clickable = !!onRowClicked;
+  const tableClass = classNames('table', { '-clickable': clickable });
 
-  const handleClick = key => {
-    let sortNext = SORT.DEFAULT;
+  const columnClass = className => (className && className()) || '';
+  const columnLabel = (label, key) => (label || label === '' ? label : capitalize(key));
 
-    if (sorting.type === SORT.DEFAULT || sorting.by !== key) {
-      sortNext = SORT.ASCENDING;
-    } else if (sorting.type === SORT.ASCENDING) {
-      sortNext = SORT.DESCENDING;
-    }
-
-    setSorting({ type: sortNext, by: key });
-  };
-
-  const handleSort = (a, b) => {
-    if (sorting.type === SORT.ASCENDING) {
-      return b[sorting.by] - a[sorting.by];
-    }
-
-    if (sorting.type === SORT.DESCENDING) {
-      return a[sorting.by] - b[sorting.by];
-    }
-
-    return 0;
-  };
-
-  useEffect(() => {
-    return () => setSorting(DEFAULT_SORTING);
-  }, [rows]);
-
-  const tableClass = classNames({ table: true, '-clickable': clickable });
+  const baseRowClass = (className, originalValue) => (className ? className(originalValue) : '');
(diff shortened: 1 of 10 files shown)


psikoi:
Oh nice! Join the discord man, we talk dev in there

app/src/components/Table/Table.scss
```diff
@@ -44,11 +44,51 @@
     font-size: 0.8em;
     font-weight: 600;
     text-align: left;
+    cursor: pointer;
 
     &.-collapse {
       font-size: 0;
       padding: 0;
     }
+
```
psikoi, line 53: Can you use css modifiers for this?

```
.arrow {
   
    &.-ascending {
       *override the default values here*
    }
   
    &.-descending{
       *override the default values here*
    }


    *default arrow attributes here*
}
```

app/src/components/Table/Table.jsx
```diff
@@ -23,17 +69,26 @@ function Table({ rows, columns, highlightedIndex, onRowClicked, clickable }) {
       <thead>
         <tr>
           {columns.map(({ key, label, className }) => {
+            const customClass = (className && className()) || '';
+            const arrowClass = `arrow-${sortData.value}`;
             return (
-              <th className={className && className()} key={`col-${key}`}>
+              <th className={customClass} key={`col-${key}`} onClick={() => handleClick(key)}>
                 {label || label === '' ? label : capitalize(key)}
+                <div className={`${customClass} arrow`}>
```
psikoi, line 77: Following my comment below about modifiers, you could then use the `classNames` package imported at the top of the file to do something like:

`
className = {"arrow", {"-ascending": isAscending, "-descending": isDescending}`

app/src/components/Table/Table.jsx
```diff
@@ -1,15 +1,61 @@
...
+    };
+
+    setSortData(data);
+  };
+
+  const handleSort = (a, b) => {
+    if (sortData.type === SORT.ASCENDING) {
+      return b[sortData.by] - a[sortData.by];
+    }
+    if (sortData.type === SORT.DESCENDING) {
```
psikoi, line 46: Mind adding a line above that second "if"?

app/src/components/Table/Table.jsx
```diff
@@ -1,15 +1,61 @@
...
+      return a[sortData.by] - b[sortData.by];
+    }
+
+    return 0;
+  };
+
+  useEffect(() => {
```
psikoi, line 53: What is this for? Is it for restoring the default sorting when the table gets unmounted?

breeku: Yeah, otherwise if you sort by example by 'rank' in skilling, then change the tab to bossing, 'rank' will be sorted by default. Not sure if that's a bad thing though.

psikoi: ah makes sense

psikoi:
I renamed some variables and added isSortable prop to table columns (there are some columns where sorting isn't needed).

However there's a bigger issue with this, the sorting only works for direct columns, for example:

```
{ 
    key: 'rank', 
    width: 70 
},
{
  key: 'start',
  transform: val => <NumberLabel value={val} />,
  className: () => '-break-large',
  get: row => (row.progress ? row.progress.start : 0)
}
```

It will work for `rank`, since the object key is "row", but for `start `for example, it has a custom "get" method, and those columns aren't being sorted, as row[start] is undefined (it should be row.progress.start).

I'll do some refactor on the Table component in general since it's getting pretty messy now, once it's cleaner, i'll fix this PR to correctly sort all columns 👍

psikoi:
- Refactored Table component
- Fixed string sorting (it wasn't sorting string values like "username")

breeku:
> 
> 
> I renamed some variables and added isSortable prop to table columns (there are some columns where sorting isn't needed).
> 
> However there's a bigger issue with this, the sorting only works for direct columns, for example:
> 
> ```
> { 
>     key: 'rank', 
>     width: 70 
> },
> {
>   key: 'start',
>   transform: val => <NumberLabel value={val} />,
>   className: () => '-break-large',
>   get: row => (row.progress ? row.progress.start : 0)
> }
> ```
> 
> It will work for `rank`, since the object key is "row", but for `start `for example, it has a custom "get" method, and those columns aren't being sorted, as row[start] is undefined (it should be row.progress.start).
> 
> I'll do some refactor on the Table component in general since it's getting pretty messy now, once it's cleaner, i'll fix this PR to correctly sort all columns 👍

That's a good catch, I somehow didn't even try the table on other than the player page. My bad...

> 
> 
>     * Refactored Table component
> 
>     * Fixed custom column "get" selector sorting
> 
>     * Fixed string sorting (it wasn't sorting string values like "username")
> 
> 
> @breeku take a look at the code and let me know if I'm missing something 👍

Did you notice these warnings
`Line 57:6:  React Hook useEffect has a missing dependency: 'onTimerEnded'. Either include it or remove the dependency array  react-hooks/exhaustive-deps` 
and
`Line 75:64:  React Hook useMemo has a missing dependency: 'handleSort'. Either include it or remove the dependency array  react-hooks/exhaustive-deps`

Otherwise looks good! From what I tested now it works well :)

PR #152, opened 2020-05-25
fix group table overflow arrow and add/fix labels

breeku:
Fixes:
```
ySomicToday at 03:03
Also on mobile, the sortable table has an arrow for me (probably from a column that's hidden duo responsive design)
```

Adds label for displayName

### DIFF
--- a/app/src/pages/Group/components/MembersTable/MembersTable.jsx
+++ b/app/src/pages/Group/components/MembersTable/MembersTable.jsx
@@ -21,6 +21,7 @@ function MembersTable({ members, isLoading }) {
       },
       {
         key: 'displayName',
+        label: 'Display name',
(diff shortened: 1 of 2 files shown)


app/src/pages/Group/components/MembersTable/MembersTable.jsx
```diff
@@ -21,6 +21,7 @@ function MembersTable({ members, isLoading }) {
       },
       {
         key: 'displayName',
+        label: 'Display name',
```
psikoi, line 24: Can you just make this "Name"?
 
Oh and while you're at it, mind checking if i missed any other tables' "displayName" column? :D

psikoi:
Thanks

PR #153, opened 2020-05-25
fix css load order

breeku:
I noticed that there was a difference in css when using prod / build.

prod:

build:

This was due to "wrong" load order of css 

prod after fix:


psikoi:
Oh wow! I've wondered what caused this, I assumed it was some bootstrap weirdness, love how simple this was 😆 Thanks

PR #189, opened 2020-06-02
Feature: TypeScript conversion [WIP]

Jakesterwars:


Jakesterwars:
This doesn't have a few updates over the past couple days since I started, so I'll have to manually get those in when I can.

Jakesterwars:
This will close #190 when it's fully merged.

Jakesterwars:
Moved away from TypeORM for a bit, reverted back to use sequelize-typescript.

PR #243, opened 2020-06-18
Fix sort reset

### LINKED ISSUE #224: Clicking a metric in a 'sorted' table selects wrong metric and reset sort
**Expected behaviour**

When sorting a table, and clicking a metric, have that metric that you clicked selected and update the graph.

**Actual behaviour**

Table sorts reset and selects the metric on the same position you clicked. Resulting in the wrong item selected

**Reproducing steps**

1. Go to a profile gained page [Example](https://wiseoldman.net/players/4029/gained/skilling). 
2. Sort a table by clicking "Exp" 
3. Click a metric that isn't in it's original position

Gif

breeku:
Closes #224

- [x] Only reset sort on metric change
- [x] Select the correct metric after sort
- [x] Highlight the correct metric after sort

### DIFF
--- a/app/src/components/Table/Table.jsx
+++ b/app/src/components/Table/Table.jsx
@@ -21,7 +21,7 @@ function Table({
   rows,
   columns,
   uniqueKeySelector,
-  highlightedIndex,
+  highlightedMetric,
   listStyle,
   onRowClicked,
   metricType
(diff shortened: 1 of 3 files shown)


psikoi:
I don't understand what this PR does, you first had Defence selected, then by sorting the table, Woodcutting became selected? Isn't it supposed to stay with Defence?

breeku:
> 
> 
> I don't understand what this PR does, you first had Defence selected, then by sorting the table, Woodcutting became selected? Isn't it supposed to stay with Defence?

Let's just say heatstroke got the better of me 🗡️ I updated the gif

PR #246, opened 2020-06-20
Server TypeScript conversion

### LINKED ISSUE #190: TypeScript conversion - Server
To have tracking for this as a whole, this will be the issue related to the server conversion to TypeScript as a whole. There will be a separate one for the App itself to move to TSX.

Jakesterwars:
Closes #190


server/src/database/index.ts
```diff
@@ -0,0 +1,49 @@
...
+  Membership,
+  Participation,
+  Player,
+  Record,
+  Snapshot
+} from './models';
+
+dotenv.config({ path: process.env.NODE_ENV === 'test' ? '.env.test' : '.env' });
+
+const sequelize = new Sequelize({
```
psikoi, line 17: Why isn't this using the configs from the config.ts on the same folder?

server/src/database/models/achievement.model.ts
```diff
@@ -0,0 +1,57 @@
...
+
+@Table(options)
+export default class Achievement extends Model<Achievement> {
+  @PrimaryKey
+  @ForeignKey(() => Player)
+  @AllowNull(false)
+  @Column({ type: DataType.INTEGER, onDelete: 'CASCADE' })
+  playerId: number;
+
+  @BelongsTo(() => Player)
```
psikoi, line 39: Can you move these relationship variables to the end of the file?

It would be neat to have all model properties on the top, and all relationships on the bottom

server/src/api/routing.ts
```diff
@@ -1,11 +1,11 @@
...
-const logger = require('./logger');
-const playerRoutes = require('./modules/players/player.route');
-const deltaRoutes = require('./modules/deltas/delta.route');
-const recordRoutes = require('./modules/records/record.route');
-const competitionRoutes = require('./modules/competitions/competition.route');
-const groupRoutes = require('./modules/groups/group.route');
+import * as express from 'express';
+import * as logger from './logger';
+import { NotFoundError } from './errors';
+import { api as playerRoutes } from './modules/players/player.route';
```
psikoi, line 4: This is a bit strange, can't you export the correct name in the first place? `export { playerRoutes }`

server/src/api/index.ts
```diff
@@ -0,0 +1,34 @@
+import * as dotenv from 'dotenv';
+import * as express from 'express';
```
psikoi, line 2: These imports are also weird, what's the difference between that and `import express from 'express'`?

server/src/api/routing.ts
Jakesterwars: I was trying to keep naming the same as possible for the actual exports, but I can shift to the routes name anyways.

server/src/database/models/achievement.model.ts
Jakesterwars: Funny enough I originally had all the relationship stuff at the bottom of the model. I only moved away because I saw many people doing this for possibly readability for a foreign key pairing.

server/src/database/index.ts
Jakesterwars: I was having some issues with TS screaming at me for something inside Sequelize about it. I just never went back to resolve it. I'll go look at it again.

server/src/api/index.ts
Jakesterwars: If I'm remembering right, it depends on how you set up you TSconfig to handle default exports that may or may not exist. I'll look at this again.

server/src/database/index.ts
Jakesterwars: Did what I could to use the config more, but the dialect is still an issue so I had to keep the function for checking env.

server/src/database/models/achievement.model.ts
Jakesterwars: Moved all the relationship stuff in the models down to the bottom.

server/src/api/index.ts
Jakesterwars: After moving to esModuleInterop to true, I went through and fixed the ones I found that did this.

Jakesterwars:
Closing in favor of #315

Issue #341, opened 2020-07-26
[API] Players "Track" API doesn't detect inexistent usernames


ldionmarcil:
When issuing a `/api/players/track` on a player that doesn't exist in the hiscores, the API returns with a validation error:
 
```2020-07-26 13:55:49,069 INFO [wom:40] Reaching https://wiseoldman.net/api/players/username/doesntexist5
2020-07-26 13:55:49,349 INFO [wom:55] Updating tracking for doesntexist5
2020-07-26 13:55:49,349 INFO [wom:40] Reaching https://wiseoldman.net/api/players/track
2020-07-26 13:55:49,350 INFO [wom:45] Request body: {'username': 'doesntexist5'}
2020-07-26 13:55:56,299 INFO [wom:70] Fetch failed for doesntexist5: {'message': "Validation error: Cannot read property 'length' of undefined"}
```

After this, the inexistent username is added to the snapshot database, which it shouldnt as they do not exist:
```2020-07-26 13:56:40,694 INFO [wom:195] Fetching wom hiscore for doesntexist5
2020-07-26 13:56:40,695 INFO [wom:40] Reaching https://wiseoldman.net/api/players/username/doesntexist5
[...]
{"id": 27332, "username": "doesntexist5", "displayName": "doesntexist5", "type": "unknown", "lastImportedAt": null, "registeredAt": "2020-07-26T17:53:41.191Z", "updatedAt": "2020-07-26T17:53:41.191Z", "latestSnapshot": null, "combatLevel": null}
```

This causes confusion on the API side, as the returned message should indicate that the player does not exist in the hiscores. Also, that username should most likely not be added to your database.

ldionmarcil:
Could the API respond with an appropriate status code for player not found, such as 404? Or, the status code for "looked up too recently" could be changed to 429, to easily identify the distinction between replies.

Thanks

psikoi:
> Could the API respond with an appropriate status code for player not found, such as 404? Or, the status code for "looked up too recently" could be changed to 429, to easily identify the distinction between replies.

Both of these error codes have been implemented. Closing.

PR #363, opened 2020-08-03
Added HiscoreNotFoundError for when player is not in Hiscores

ldionmarcil:
Changes for comments made in #341


psikoi:
Error codes 404 have been added to all player/group/competition fetches. Also "Updated too soon" is now a 429 error code. Closing.

PR #465, opened 2020-11-08
Add confirmation modal when removing group members or competition participants

### LINKED ISSUE #360: Add remove confirmation modal to group/competitions
When a group member or competition participant gets removed, a modal should appear asking for confirmation before submitting the form, showing which players are going to be removed.

lukaszklis:
Fixes #360

### DIFF
--- a/app/src/pages/EditCompetition/EditCompetition.jsx
+++ b/app/src/pages/EditCompetition/EditCompetition.jsx
@@ -15,8 +15,8 @@ import ParticipantsSelector from '../../components/ParticipantsSelector';
 import ImportPlayersModal from '../../modals/ImportPlayersModal';
 import { getMetricIcon, getMetricName } from '../../utils';
 import { ALL_METRICS } from '../../config';
-import './EditCompetition.scss';
 import RemoveParticipantsModal from '../../modals/RemoveParticipantsModal/RemoveParticipantsModal';
+import './EditCompetition.scss';
 
 function getMetricOptions() {
   return ALL_METRICS.map(metric => ({
--- a/app/src/pages/EditCompetition/EditCompetition.jsx
+++ b/app/src/pages/EditCompetition/EditCompetition.jsx
@@ -124,8 +124,6 @@ function EditCompetition() {
 
   const handleRemoveParticipantsModalClose = () => {
     hideRemoveParticipantsModal();
-    setParticipants(currentParticipants => [...currentParticipants, ...removedParticipants]);
-    setRemovedParticipants([]);
   };
 
   const handleVerificationCodeChanged = e => {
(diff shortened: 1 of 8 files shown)


psikoi:
This needs to be rebased, caused by #473

lukaszklis:
Will take care of it later today!

app/src/pages/EditCompetition/EditCompetition.jsx
```diff
@@ -18,6 +18,7 @@ import fetchDetailsAction from '../../redux/modules/competitions/actions/fetchDe
 import editAction from '../../redux/modules/competitions/actions/edit';
 import { getCompetition, isEditing, getError } from '../../redux/selectors/competitions';
 import './EditCompetition.scss';
+import RemoveParticipantsModal from '../../modals/RemoveParticipantsModal/RemoveParticipantsModal';
```
psikoi, line 21: Can this be moved to the line above it? I like to keep the scss as the last import

app/src/pages/EditGroup/EditGroup.jsx
```diff
@@ -14,6 +14,7 @@ import fetchDetailsAction from '../../redux/modules/groups/actions/fetchDetails'
 import fetchMembersAction from '../../redux/modules/groups/actions/fetchMembers';
 import { getGroup, isEditing, getError } from '../../redux/selectors/groups';
 import './EditGroup.scss';
+import RemoveGroupMembersModal from '../../modals/RemoveGroupMembersModal';
```
psikoi, line 17: Can this be moved above the scss import?

psikoi:
Did some testing on this.

Initial group members: `Psikoi`, `Lynx Titan`, `Zezima`

Step 1: Removed `Lynx Titan`
Step 2: Tried to submit the form, and was correctly informed that I would be deleting `Lynx Titan`
Step 3: Added `lynx titan`
Step 4: Removed `lynx titan` and `Zezima`
Step 2: Tried to submit the form, and was **incorrectly** informed that I would be deleting `Lynx Titan`, `lynx titan` and `Zezima`

TLDR: `lynx titan` was never a part of the group (since the form wasn't submitted) but yet the name showed up on the modal.

I think the cleanest solution (that would remove a lot of logic from this) is to simply:

When submitting the form, compare the updated members list with the original members list, and pick out the ones missing from the updated list (meaning they were removed). This would prevent the issue mentioned previously, as `lynx titan` wouldn't be in the original members list.

psikoi:
Also: 
> Are you sure you want to remove following members from this group?
Could probably be improved to
> Are you sure you want to remove **the** following members from this group?

lukaszklis:
Closing in favor of #482 (where all of the feedback from this PR has been implemented).

PR #468, opened 2020-11-08
Improve click area for dropdowns in navigation bar

### LINKED ISSUE #432: Improvement: Dropdown menus
Current in the top navigation bar where you can drop down leaderboards and community you need to click text to go to the new page. This should be improved to be the entire dropdown "card" instead for ease of use.

lukaszklis:
Fixes #432


psikoi:
Thanks!

PR #482, opened 2020-11-19
Add confirmation modal for removing group members/participants

### LINKED ISSUE #360: Add remove confirmation modal to group/competitions

lukaszklis:
Fixes #360.


lukaszklis:
As discussed with @psikoi, we’ll merge the two modals into a more generic one `RemovePlayersModal` and rename `utils/users.js` to `utils/players.js`

psikoi:
I think those new functions in the player.js function was a bit too far into abstraction, but everything else looks good to go 😄

psikoi:
Added a commit with some code style changed + the username standardization fix:

Changed the mapMembers function to be the inner function of the map, instead of the map itself, I believe it makes it easier to understand what the code might do (Also moved the mapMember function to outside of the component's render)


Changed the removed players calc a bit, to make it easier to read, also removed the double mapping (mapping with `mapMembers `before filtering, and then mapping again to only the `displayName `on `setRemovedPlayers`)

lukaszklis:
@psikoi I‘be tested it locally too and it seemed to be working just fine. I’d say it is good to go for the next release

PR #502, opened 2020-11-28
Various alias fixes, add skill aliases

Sethmare:
-Added aliases for all skills
-Removed spaces from existing aliases
-Add a few new aliases for other metrics


psikoi:
Some of these a bit unnecessary, do people really say `r/c` or `w/c`?

Also `stronk`, `chop`, `archery`, etc etc.

I understand it might be similar to the actual skill name, but I'd rather not go down thisroad, Skill names are pretty easy to memorize.

PR #535, opened 2020-12-05
Allow scrolling through remove players modal

lukaszklis:
Fixes a problem reported on Discord earlier today which prevents users from confirming removal of players from groups/competitions with a lot of members.

Preview:
Dec-06-2020 11-10-14


psikoi:
Can you change it so that only the names are scrollable? making the submit button always visible

lukaszklis:
I had that variant earlier however I found it a bit confusing. Will share some details on the #development channel on Discord and we can take a decision on how to fix this.

PR #543, opened 2020-12-15
Show total ehp under EHP instead of Experience

rorro:
The total EHP is currently shown under the Experience category in the Gains section. This puts it under the EHP category instead which also makes it sortable correctly.

Before
before
After
after

EHP and EHB were also not rounded in the Overview. Currently shows 5 decimals, this rounds it to 2 decimals like everything else.


psikoi:
O.o I'm pretty sure I fixed this recently

psikoi:
Nvm, I only changed it for the stats table, not the gained table.

Referencing #494

PR #544, opened 2020-12-15
Update docs

### LINKED ISSUE #426: Incorrect documentation on edit group's endpoint
The example given as request body indicates parametter members as a list of string but it only works if a list of objects is given.

rorro:
I will go through them all but all of them might not need updating.

- [x] Achievements
- [x] Competitions
- [x] Deltas
- [x] Groups
- [x] Names
- [x] Players
- [x] Records
- [x] Snapshots

Closes #426

### DIFF
--- a/docs/src/configs/docs/deltas/entities.js
+++ b/docs/src/configs/docs/deltas/entities.js
@@ -445,12 +445,12 @@ export default [
   {
     name: 'Player types',
     description:
-      'All the possible values for the "playerType" field of the delta model.',
+      'All the possible values for the "playerType" query parameter.',
     values: ['unknown', 'regular', 'ironman', 'hardcore', 'ultimate']
   },
   {
     name: 'Player Builds',
-    description: 'All the possible values for the "build" field of the delta model.',
+    description: 'All the possible values for the "playerBuild" query parameter.',
     values: ['main', '1def', 'lvl3', '10hp', 'f2p']
   },
   {
--- a/docs/src/configs/docs/deltas/entities.js
+++ b/docs/src/configs/docs/deltas/entities.js
@@ -443,6 +443,12 @@ export default [
     values: ['6h', 'day', 'week', 'month', 'year']
   },
   {
+    name: 'Player types',
+    description:
+      'All the possible values for the "playerType" field of the delta model.',
+    values: ['regular', 'ironman', 'ultimate', 'hardcore']
+  },
+  {
     name: 'Metrics',
     description: 'All the possible values for the "metric" field of the leaderboard endpoint.',
     values: [
@@ -524,7 +530,9 @@ export default [
       'vorkath',
       'wintertodt',
       'zalcano',
-      'zulrah'
+      'zulrah',
+      'ehp',
+      'ehb'
     ]
   }
 ];
--- a/docs/src/configs/docs/deltas/entities.js
+++ b/docs/src/configs/docs/deltas/entities.js
@@ -446,7 +446,12 @@ export default [
     name: 'Player types',
     description:
       'All the possible values for the "playerType" field of the delta model.',
-    values: ['regular', 'ironman', 'ultimate', 'hardcore']
+    values: ['unknown', 'regular', 'ironman', 'hardcore', 'ultimate']
+  },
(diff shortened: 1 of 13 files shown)


rorro:
Things to take a second look at.

When removing a player from a group. Is this message possible or not.
```
{
  description: 'If none of the members given exist.',
  body: { message: 'No valid tracked players were given.' }
}
```

psikoi:
Can you replace the base URL `https://wiseoldman.net/api` with `https://api.wiseoldman.net` in the API introduction page please?

psikoi:
Also on that same page, the rate limit isn't 200 anymore, it is now 500

docs/src/configs/docs/names/endpoints.js
```diff
@@ -60,23 +73,35 @@ export default [
...
       {
-        id: 12,
-        playerId: 13432,
-        status: 0,
-        oldName: 'zezima',
-        newName: 'lynx titan',
-        updatedAt: '2020-08-05T23:58:04.194Z',
-        createdAt: '2020-08-05T23:58:04.194Z',
-        resolvedAt: null
+        description: 'Note: Only showing top 3 for demo purposes',
```
psikoi, line 76: This note doesn't apply to this endpoint, could be a copy & paste mistake?

docs/src/configs/docs/records/entities.js
```diff
@@ -42,6 +42,18 @@ export default [
       'All the possible values for the "period" field of the record model. (Note: a month is 31 days)',
     values: ['6h', 'day', 'week', 'month', 'year']
   },
+  {
+    name: 'Player types',
+    description:
+      'All the possible values for the "playerType" field of the record model.',
```
psikoi, line 48: There is no playerType in the record model, maybe change this to `...for the "playerType" query parameter.`

**The same applies for the playerBuild below.**

docs/src/configs/docs/deltas/entities.js
```diff
@@ -442,6 +442,17 @@ export default [
       'All the possible values for the "period" field of the delta model. (Note: a month is 31 days)',
     values: ['6h', 'day', 'week', 'month', 'year']
   },
+  {
+    name: 'Player types',
+    description:
+      'All the possible values for the "playerType" field of the delta model.',
```
psikoi, line 448: There is no playerType in the delta model, maybe change this to ...for the "playerType" query parameter.

The same applies for the playerBuild below.

psikoi:
`JsonBlock.jsx` line 5:

If you add `white-space: pre-wrap` to those styles, it wraps the json text and resolves that issue with the long error messages.

docs/src/configs/docs/records/entities.js
rorro: Fixed in ae1a27c

docs/src/configs/docs/deltas/entities.js
rorro: Fixed in ae1a27c

psikoi:
`Participation` in competition/entities is missing two properties:

- startSnapshotId
- endSnapshotId

These are the first and last snapshots of a player's competition participation. The player's progress is the subtraction (endSnapshot - startSnapshot)

psikoi:
In "Create Competition", there's the following comment:

> The response will contain a "verificationCode", this code must be stored as it is not possible to edit or delete the competition at a later date without it.

Maybe it could be edited to be

> If not a group competition, the response will contain a "verificationCode", this code must be stored as it is not possible to edit or delete the competition at a later date without it.

psikoi:
On competition "update all":

> This action will perform a soft-update, meaning it won't import the player from CML or determine it's type.

This comment is outdated and no longer true

psikoi:
On "GET/groups/:id/gained" it seems the success response has 1 too many levels of [ ]

psikoi:
Everything else looks good to go! 👍

rorro:
> On "GET/groups/:id/gained" it seems the success response has 1 too many levels of [ ]

rorro:
I did not mean to close it. Oh well. The above comments have been fixed in 26b3abb

PR #610, opened 2021-01-10
Export group members list

Jakesterwars:
Definitely not the second PR for this, false alarm.

* Allows the user to export the current group's member name list into a CSV file.


app/src/pages/Group/Group.jsx
```diff
@@ -6,6 +6,7 @@ import { useUrlContext } from 'hooks';
 import { Loading, Tabs } from 'components';
 import { groupActions, groupSelectors } from 'redux/groups';
 import { competitionActions } from 'redux/competitions';
+import saveCsv from 'save-csv';
```
psikoi, line 9: Can you move this import to above all the local imports? under "react-helmet"

Jakesterwars: Movie groovied

PR #649, opened 2021-02-03
home-page: add counters

dekvall:
image


AllanOcelot:
Love the idea, looks like there's an ESlint error on the build though

psikoi:
Yes there's something wrong with our eslint version

PR #750, opened 2021-04-28
use group info endpoint for temple migrations

dekvall:


server/src/api/services/external/temple.service.ts
```diff
@@ -1,21 +1,25 @@
...
     const { data } = await axios.get(URL);
 
-    if (!data || !data.length) {
+    if (!data) {
       throw new Error();
     }
 
-    return data;
+    const { members, leaders } = data.data;
+    const { name } = data.data.info;
```
psikoi, line 20: this could be changed to:

```
    const { members, leaders, info } = data.data;
    const { name } = info;
```

server/src/api/services/external/cml.service.ts
```diff
@@ -45,10 +46,12 @@ async function fetchGroupMembers(gid: number): Promise<string[]> {
       throw new Error();
     }
 
-    return players.split('\n').filter(n => !!n);
+    const members = players.split('\n').filter(n => !!n);
+
+    return { members: members };
```
psikoi, line 51: This could just be `return { members };`

PR #800, opened 2021-05-30
#742 show start and end comp times in UTC as well as local

### LINKED ISSUE #742: Show an extra UTC date on competition pages
Currently, the "Starts At" / "Ends at" fields in the competition page display local time. It could be good if could add two other fields that would display those dates as UTC.

aleccaputo:
Closes #742


aleccaputo:
lint will be fixed with #802

app/src/utils/dates.js
```diff
@@ -1,13 +1,19 @@
 import moment from 'moment';
 
+const defaultDateMask = 'MM-DD-YYYY HH:mm';
```
psikoi, line 3: small nitpick, but could you make this constant's name all uppercase?

PR #801, opened 2021-05-31
#715 Add zerker player build

### LINKED ISSUE #715: Add zerker player build
Requirements: 45 def and 52 prayer

The changes would be essentially the same as https://github.com/wise-old-man/wise-old-man/pull/392 but with different name and requirements.

aleccaputo:
Closes #715


app/src/utils/player.js
```diff
@@ -34,6 +34,8 @@ export function getPlayerBuild(build) {
       return 'F2P';
     case '10hp':
       return '10 Hitpoints Pure';
+    case 'zerker':
+      return 'Berserker';
```
psikoi, line 38: Code looks good, just wondering if "Berserker" is really the way we should call them, rather than "Zerker" 🤔

aleccaputo: Changed

Issue #819, opened 2021-06-17
Fix player username filtering


psikoi:
Apparently it's possible to add players with special characters in their usernames via the group/add-member endpoint.

Best thing to do is ensure every usage of `playerService.findAllOrCreate` has some protection, by using `playerService.standardize`.

aleccaputo:
@psikoi it looks like it already does that
`async function findAllOrCreate(usernames: string[]): Promise<Player[]> {
  const foundPlayers = await findAll(usernames);

  if (foundPlayers.length === usernames.length) return foundPlayers;

  const foundUsernames = foundPlayers.map(f => f.username);
  const standardizedUsernames = usernames.map(standardize);
  const missingUsernames = standardizedUsernames.filter(u => !foundUsernames.includes(u));

  const newPlayers = await Player.bulkCreate(
    missingUsernames.map(m => ({ username: standardize(m), displayName: sanitize(m) })),
    { individualHooks: true }
  );

  return [...foundPlayers, ...newPlayers].sort(
    (a, b) => standardizedUsernames.indexOf(a.username) - standardizedUsernames.indexOf(b.username)
  );
}`

psikoi:
Fixed in #820

PR #820, opened 2021-06-17
#819 don't allow invalid usernames to be added to group

aleccaputo:


psikoi:
Can you also check if addParticipants and addTeams in competition service need the same handling please? Code looks good 👍

PR #830, opened 2021-06-24
Feature: Add new clan roles/ranks

psikoi:
⚠️ Note to self: Don't forget to manually edit the memberships database table


fatfingers23:
Hey. I didn't see an issue covering the clan implementation. I notice there's no api that exports a list of OSRS clanmate names like they do in rs3 on the hiscore api. So I made a runelite plugin to do it. Wanted to share that here if there was issues with getting clanmate data on clans. It has the ability to add a url end point for a post.

https://runelite.net/plugin-hub/show/clanmate-export

PR #846, opened 2021-07-25
#842 case insensitive search

### LINKED ISSUE #842: Make name change search case insensitive
Currently searching the name changes is case sensitive, causing any capital letters in the search bar to pull up no results since most(if not all) usernames are currently lowercase. This should be changed to be case insensitive to show more accurate results

aleccaputo:
closes #842


aleccaputo:
found this in the docs under operators. Let me know if this is a good way to do this
https://sequelize.org/v3/docs/querying/

psikoi:
Looks good to me! That operator only works on postgres, and our integration tests run on SQLite, but i haven't written tests for name changes, so it's a problem for another day.

Thanks!

PR #866, opened 2021-08-16
Fix spelling of clan title Marshal

rorro:


psikoi:
Probably would need to change the icon's file name too right?

rorro:
It seems like a file with the correct name already exists.
Marshal: https://github.com/wise-old-man/wise-old-man/blob/master/app/public/img/runescape/roles/Marshal.png
Marshall: https://github.com/wise-old-man/wise-old-man/blob/master/app/public/img/runescape/roles/Marshall.png

psikoi:
Weird, can you remove the unused one, then?

rorro:
Done

PR #881, opened 2021-09-02
#808 - Change Base X Stats Percentage Calculation

### LINKED ISSUE #808: Base X stats achievement progress incorrectly calculated
The achievement partial progress towards base X stats is improperly calculated. On an account with all except 3 skills at 90+ (and those skills are 88+), the achievement progress reads "Base 90 Stats 55.22% completed". The account is probably more accurately 90%+ of the way to base 90s. Other base X achievement partial progress may also be improperly calculated, although there seems to be no issue with actually awarding the achievement when base X is reached.

jkinequon:
The issue closes #808 
**Problem**
The issue is that the previous solution calculates the percentage of base stats being
`PROGRESS_TO_90_BASE = LOWEST_SKILL'S_EXP / EXP_AT_LVL_90` which doesn't give a true accuracy of how far you are to the Base stats that is being calculated.

**Solution**
We are changing the logic from this equation to this new one being `PROGRESS_TO_90_BASE = SUM_OF_ALL_SKILLS_CAPPED_AT_90 / EXP_AT_LVL_90 * TOTAL_SKILLS`

I went ahead and based the total length of skills based on the real skills so that it automatically determines the length even when a new skill is added. Because of how the achievement progresses were coded, I had to implement some logic to make sure that the solution works.

### DIFF
--- a/app/src/pages/Player/components/AchievementGroup/AchievementGroup.jsx
+++ b/app/src/pages/Player/components/AchievementGroup/AchievementGroup.jsx
@@ -2,6 +2,7 @@ import React, { Fragment } from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import { getMetricIcon, getMetricName, formatDate, formatNumber, getLevel } from 'utils';
+import { TOTAL_SKILLS } from 'config';
 import './AchievementGroup.scss';
 
 function AchievementGroup({ group, metricType }) {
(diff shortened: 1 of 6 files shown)


app/src/pages/Player/components/AchievementGroup/AchievementGroup.jsx
```diff
@@ -92,6 +93,10 @@ function formatThreshold(threshold) {
     return getLevel(threshold + 100).toString();
   }
 
+  if ([273742 * TOTAL_SKILLS, 737627 * TOTAL_SKILLS, 1986068 * TOTAL_SKILLS, 5346332 * TOTAL_SKILLS, 13034431 * TOTAL_SKILLS].includes(threshold)) {
```
psikoi, line 96: `[273742 * TOTAL_SKILLS, 737627 * TOTAL_SKILLS, 1986068 * TOTAL_SKILLS, 5346332 * TOTAL_SKILLS, 13034431 * TOTAL_SKILLS]`

could probably be rewritten as

`[273742, 737627, 1986068, 5346332, 13034431].map(i => i * TOTAL_SKILLS)`

to look a little cleaner

psikoi: And if that's the case, should probably remove the `TOTAL_SKILLS` and `REAL_SKILLS` as they're now only used once, and instead write that map as 

`[273742, 737627, 1986068, 5346332, 13034431].map(i => i * SKILLS.filter(s => s !== 'overall').length)`

psikoi (commented):
Wrote these comments on the app-side of the PR, but I believe it also applies to the server-side 👍

PR #952, opened 2022-07-15
Add additional statistics to the group statistics tab.

### LINKED ISSUE #940: Suggestion: Additional stats on the Group Statistics page
[](https://cdn.discordapp.com/attachments/696219233167736884/958074411657887794/unknown.png)I feel we could add some additional stats on the Groups Statistics page.

- Average Combat level
- Average EHP & Average EHB
- Amount of members with an Infernal cape
- Amount of members with a 200m is maybe more interesting/useful than Number of individual 200ms? At the very least I think it would be good to have it as an additional stat if you'd like to keep Number of 200ms.

Maybe even a breakdown of player type.
Number of Irons/Hardcores/Uims/Other. I say other because Regular includes GIM 🤷

I'd love to see more than just the Average statistics also. Total & Highest would both be great additions to most metrics. The rank column can safely be removed imo.

llebc:
Hi 👋,

This is my first PR so please let me know if something is incorrect and / or better way of doing things.

This is off the back of a request from @BoomOSRS.

Adds the following stats:

- Average Combat level
- Average EHP & Average EHB
- Amount of members with an Infernal cape
- Amount of members with a 200M rather than individual 200M statistics.

Also adds the following columns to the table:
- Highest value per metric - formats levels.
- The total per metric.

Also as a part of this as they are no longer average stats within this list:
- Renamed Average member stats -> Member Stats
- Renamed column: Level -> Avg Level
- Renamed column: Value -> Avg Value

Adds additional helper methods:
- `hasInfernoCapes` -> Determines if a snapshot has inferno cape by getting the count of zuk kills.
- `hasMaxedSkill` -> Determines if a snapshot has a maxed skill.

Closed #940.

### DIFF
--- a/app/src/pages/Group/containers/Statistics.jsx
+++ b/app/src/pages/Group/containers/Statistics.jsx
@@ -48,7 +48,7 @@ function Statistics() {
           </b>
         </div>
         <div className="statistics-block">
-          <span className="statistic-label">Number of a 200ms</span>
+          <span className="statistic-label">Total 200ms</span>
           <b className="statistic-value">
             {showPlaceholder ? 'Loading...' : statistics.maxed200msCount}
           </b>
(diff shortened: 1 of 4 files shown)


psikoi:
this PR is stale now, closing

arvcork:
Will be picking this one up as part of a conversation I have had with @psikoi  on Discord in the past. 

If possible, can this be re-opened and assigned to me please?

Issue #939, opened 2022-03-28
Update-all on comps - error message refers to 1hr cooldown instead of 24hr.


BoomOSRS:
https://discord.com/channels/679454777708380161/705827679538249738/930584515754225687
As per above discord anouncement, the cooldown on update-alls on competitions should be 24hrs if not within the first or last 6hrs of the competition. If attempting to update-all on a competition when no users are oudated (updated over 24hrs ago) the error message still refers to the old 1h cooldown.

![image](https://user-images.githubusercontent.com/70453840/160459090-cea31633-e31c-48bc-9e1f-f0bbfd87fcaf.png)

psikoi:
Fixed in #980

PR #980, opened 2022-09-25
Correct error message about "Update All" cooldown

Sethmare:


Sethmare:
Should close #939

psikoi:
I think this is wrong actually, competition update alls aren't always 24h, they can be 1h cooldown if in the first or last 6h of a competition.

so this should be something like 

```
throw new BadRequestError(`This competition has no outdated participants (updated over ${cooldownDuration}h ago).`);
```

`cooldownDuration` can either be `1` or `24` depending on the conditions i explained above

PR #1025, opened 2022-10-23
Change Base X Achievement Calculation

### LINKED ISSUE #808: Base X stats achievement progress incorrectly calculated

GoryMoon:
This is an update of #881 that fixes #808.

This is basically the same as the previous PR, now updated for the changes since the last PR was made.
Also applied comments from the other PR.

### DIFF
--- a/app/src/pages/Player/components/AchievementGroup/AchievementGroup.jsx
+++ b/app/src/pages/Player/components/AchievementGroup/AchievementGroup.jsx
@@ -1,7 +1,7 @@
 import React, { Fragment } from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
-import { MetricProps, getLevel } from '@wise-old-man/utils';
+import { MetricProps, getLevel, SKILLS } from '@wise-old-man/utils';
 import { getMetricIcon, formatDate, formatNumber } from 'utils';
 import './AchievementGroup.scss';
 
@@ -89,8 +89,9 @@ function formatThreshold(threshold) {
     return threshold;
   }
 
-  if ([273742, 737627, 1986068, 5346332, 13034431].includes(threshold)) {
-    return getLevel(threshold + 100).toString();
+  const totalSkills = SKILLS.filter(s => s !== 'overall').length;
+  if ([273742, 737627, 1986068, 5346332, 13034431].map(i => i * totalSkills).includes(threshold)) {
+    return getLevel((threshold / totalSkills) + 100).toString();
   }
 
   if (threshold <= 10000) {
--- a/app/src/pages/Player/components/AchievementGroup/AchievementGroup.jsx
+++ b/app/src/pages/Player/components/AchievementGroup/AchievementGroup.jsx
@@ -1,7 +1,7 @@
 import React, { Fragment } from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
-import { MetricProps, getLevel, SKILLS } from '@wise-old-man/utils';
+import { MetricProps, getLevel, REAL_SKILLS } from '@wise-old-man/utils';
 import { getMetricIcon, formatDate, formatNumber } from 'utils';
 import './AchievementGroup.scss';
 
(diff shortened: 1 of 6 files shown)


GoryMoon:
Rebased on the latest version changes.
@psikoi can you take a look at this

Here is an image of the change: 
WiseOldManChange

GoryMoon:
Have you had the time to check this yet? @psikoi

psikoi:
@GoryMoon sorry for the huge delay, this PR at a time when I was fully focused on release v2 of our API and I overlooked it (and others too)

I'm not sure about this implementation, I don't understand why you need to modify `getAchievementName` and the `thresholds` in the templates file.


`getCurrentValue` is a function that is expected to return 0-1 (percentage)
`getCappedExp` takes in a player's stats and a maximum exp value, and returns their combined total exp with that maximum applied to every skill's exp

So if you have a player's stats and the threshold, you can turn that into a simpler expression without altering anything else:

```
% = capped(stats, THRESHOLD) / (THRESHOLD * NUMBER_OF_REAL_SKILLS)


Examples:

// if a player has > 90 in everything except 80 in hunter
x = ((22 * 5_346_332) + 1_986_068) / (5_346_332 * 23)
x = 0,9727

// if a player has > 90 in 10 skills, and 80 in all others
x = ((10 * 5_346_332) + (13 * 1_986_068)) / (5_346_332 * 23)
x = 0,6448
```

But, to be fair, I haven't tested it so you might have some context that I'm missing. Thoughts?

GoryMoon:
`getCurrentValue` is not a value between 0-1, it's the actually current value of the skill.
It's used in https://github.com/wise-old-man/wise-old-man/blob/899d9faede1f0c5c77357519e021e210e6bb8a19/server/src/api/modules/achievements/services/FindPlayerAchievementProgressService.ts#L50-L51 to get the `absoluteProgress` and `relativeProgress`, they are in return between 0 and 1.
It's then used to show the user what the current exp of the skill is the hovering the progress. 

The `thresholds` are also used in that function to create the correct percentages.
The need for the change in `getAchievemenName` is to adjust the `threshold` before formatting to show the levels correctly. It could be moved into `formatThreshold` to where is checks the 60/70/80/90/99 thresholds.

psikoi:
oh you are correct, apologies, let me test and re-review with that in mind 🤦

server/src/api/modules/achievements/achievement.utils.ts
```diff
@@ -21,6 +22,9 @@ function getAchievementMeasure(metric: Metric, threshold: number): string {
 }
 
 function getAchievemenName(name: string, threshold: number): string {
+  if (name === 'Base {level} Stats') {
+    threshold = threshold / REAL_SKILLS.length;
```
psikoi, line 26: Could you do this same thing without assigning a new value to the parameter? maybe something like

```javascript
const adjustedThreshold = name === 'Base {level} Stats' ? threshold / REAL_SKILLS.length : threshold
```

app/src/pages/Player/components/AchievementGroup/AchievementGroup.jsx
```diff
@@ -89,8 +89,9 @@ function formatThreshold(threshold) {
     return threshold;
   }
 
-  if ([273742, 737627, 1986068, 5346332, 13034431].includes(threshold)) {
-    return getLevel(threshold + 100).toString();
+  const totalSkills = SKILLS.filter(s => s !== 'overall').length;
```
psikoi, line 92: you can import `REAL_SKILLS` from `@wise-old-man/utils` and use that instead of filtering for != overall

psikoi:
Also if you rebase on master, it'll get rid of that failing "Check App Code Quality" job

psikoi:
Pushed a commit to add a test case for this type of achievement and bumped the API version to v2.1.43, merging soon

psikoi:
tyvm @GoryMoon ! should be live in 3-4 minutes

PR #1068, opened 2023-02-16
Fix bug with Get Name Change Details that has invalid new name

Jonxslays:
This PR fixes a bug where we would attempt to read properties of null.
Inside `computePlayerMetrics` when the snapshot was null `getKillCountMap` would fail with:
`Cannot read properties of null (reading 'abyssal_sireKills')`

This occurred in the case when the new username was not tracked by WOM, AND also not on the hiscores, AND the name change request is pending.

In addition to that, make sure we only attempt to delete the `newStats.playerId` if `newStats` is not null.

Added a test to confirm this niche case and updated the documentation to note that `newStats` is now nullable and may not include id/playerid if the new username isn't tracked by WOM.


Jonxslays:
Failing tests look like a network issue in the github container or something? It's unrelated to the changes I made and all tests pass locally.

psikoi:
> Failing tests look like a network issue in the github container or something? It's unrelated to the changes I made and all tests pass locally.

it can happen, usually it works on re-run, I'll try

PR #1078, opened 2023-02-20
Only update competition participants if an array was supplied

### LINKED ISSUE #1077: Editing a competition can remove all participants
## Summary

When editing a competition by making a PUT request to `/competitions/{id}`, if no participants are supplied in the payload all participants are removed from the competition.

## Reproduction steps

- Create a new competition with at least 1 participant
- Hit the Edit Competition endpoint @ `PUT /competitions/{id}`
  - Only edit the title or any other field except participants
  - Don't include participants in the request body
- Check the participants in the competition after the request

## Expected results

Participants will be unchanged.

## Actual results

All participants are removed.

Jonxslays:
This PR fixes a bug where Edit Competition would reset participants even if the consumer did not supply a value for participants.

Closes #1077

### DIFF
--- a/server/__tests__/suites/integration/competitions.test.ts
+++ b/server/__tests__/suites/integration/competitions.test.ts
@@ -1075,6 +1075,34 @@ describe('Competition API', () => {
       expect(onParticipantsJoinedEvent).not.toHaveBeenCalled();
     });
 
+    it('should not edit teams (teams undefined)', async () => {
+      const getResponse = await api.get(`/competitions/${globalData.testCompetitionEnding.id}`);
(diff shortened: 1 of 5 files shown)


psikoi:
How does this behave when `null` is passed instead of undefined?

Jonxslays:
> How does this behave when `null` is passed instead of undefined?

null and undefined will essentially work in the same way because they are both falsy. an empty array or filled array will both be truthy, and thus trigger the participations if block in the execute function.

I'm just not a fan of undefined and prefer explicit nulls.

PR #1089, opened 2023-02-24
Add MetricLeaders to group statistics

### LINKED ISSUE #1086: addition to FetchGroupStatistics: best object
Enables the possibility to expose the best player for all metrics (skills and bosses) in a group. 

Change:
`{
  maxedCombatCount: ...,
  maxedTotalCount: ...,
  maxed200msCount: ...,
  averageStats: ...,
}` 
To: 
`{
  maxedCombatCount: ...,
  maxedTotalCount: ...,
  maxed200msCount: ...,
  averageStats: ...,
  best: {
    skills: {
      attack: {
        player: {...a player object }
        experience: ...,
        rank: ....,
        level: ....,
      },
      (other skills, etc)
    },
    bosses: {
       zulrah: {
         player: {...a player object }
         kills: ...,
         rank: ....,
       }
    },
      (other bosses, etc)
  }
}`

- Add method to snapshot.utils that can produce the above addition by finding the highest value from provided snapshots for each metric.
- Add best to return object.

Jonxslays:
This PR adds a few new things:

- The `createdAt` date for the average stats on the group statistics will now be set to the current date.
- A `MetricLeaders` object. which is basically the same as `SnapshotDataValues`, but with the addition of a `Player` for each metric.
- `GroupStatistics` now have a new field `metricLeaders` which is the metric leaders object.
- Utility functions for building the metric leaders object, and assigning players to it.
- Fixed a bug in the `average` function in snapshot utils where it would fail to throw a server error.
- Added tests.
- Documentation updates.

I'm not 100% sold on the names I went with so definitely let me know if you have better ideas.

I fetch all the players from the db once, to reduce db stress, pass the players and the snapshots into the utility function.
From there it iterates the metrics and find the best one, attaches the player, and done.

Bumped the api version to 2.2 since I would consider this a new feature, let me know if it should just be a patch though and I can update it.

Closes #1086 

EDIT:
Updated above additions for easily adding to the changelog.

### DIFF
--- a/server/src/api/modules/groups/services/FetchGroupStatisticsService.ts
+++ b/server/src/api/modules/groups/services/FetchGroupStatisticsService.ts
@@ -11,7 +11,8 @@ import {
   average,
   getCombatLevelFromSnapshot,
   getTotalLevel,
-  getMetricLeaders
+  getMetricLeaders,
+  assignPlayersToMetricLeaders
 } from '../../snapshots/snapshot.utils';
 import { GroupStatistics } from '../group.types';
 
@@ -61,7 +62,11 @@ async function fetchGroupStatistics(payload: FetchGroupStatisticsParams): Promis
   });
 
   const averageStats = format(averageSnapshot, averageEfficiencyMap);
-  const metricLeaders = getMetricLeaders(snapshots);
+
+  const { metricLeaders, leaderIdMap } = getMetricLeaders(snapshots);
+  const leaderIds = [...new Set(leaderIdMap.values())];
+  const players = await playerServices.findPlayers({ ids: leaderIds });
+  assignPlayersToMetricLeaders(metricLeaders, leaderIdMap, players);
 
   return { maxedCombatCount, maxedTotalCount, maxed200msCount, averageStats, metricLeaders };
 }
--- a/server/src/api/modules/groups/services/FetchGroupStatisticsService.ts
+++ b/server/src/api/modules/groups/services/FetchGroupStatisticsService.ts
@@ -11,7 +11,7 @@ import {
   average,
   getCombatLevelFromSnapshot,
   getTotalLevel,
-  getBestInEachMetric
+  getMetricLeaders
 } from '../../snapshots/snapshot.utils';
 import { GroupStatistics } from '../group.types';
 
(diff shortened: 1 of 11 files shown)


Jonxslays:
I noticed the tests timeout kept happening on a particular test, increased the timeout on that one from the default 5 seconds to 10 seconds. Sometimes it would come in under 5 and sometimes it would time out.
image

server/src/api/modules/groups/services/FetchGroupStatisticsService.ts
```diff
@@ -59,8 +61,10 @@ async function fetchGroupStatistics(payload: FetchGroupStatisticsParams): Promis
   });
 
   const averageStats = format(averageSnapshot, averageEfficiencyMap);
+  const players = await playerServices.findPlayers({ ids: snapshots.map(s => s.playerId) });
```
psikoi, line 64: I think it would be more efficient to

- calculate best values from everyone's snapshots
- fetch only the players that are included in the "best" list
- "insert" them into their places in the best object

In a group of 700 members, if only 50 take up the top spots, no point in fetching the other 650

server/src/api/modules/snapshots/snapshot.utils.ts
```diff
@@ -204,6 +210,87 @@ function average(snapshots: Snapshot[]): Snapshot {
...
+
+  if (!players || players.length !== snapshots.length) {
+    throw new ServerError('Invalid players list. Failed to find best players.');
+  }
+
+  return {
+    skills: Object.fromEntries(
+      SKILLS.map(s => {
+        const valueKey = getMetricValueKey(s);
+        const snapshot = [...snapshots].sort((x, y) => y[valueKey] - x[valueKey])[0];
```
psikoi, line 229: will this .sort throw if length=1 ?

Jonxslays: Don't think so.
image

server/src/api/modules/groups/services/FetchGroupStatisticsService.ts
Jonxslays: Thats very true, and here I thought I was being clever just fetching the players once XD

Jonxslays: Resolved in latest push.

psikoi:
looks good to me, merging after tests pass 👍

PR #1096, opened 2023-03-03
Fix ordering of competitions on group page

### LINKED ISSUE #796: Ordering of finished competitions on Group page to be newest-oldest
Finished competitions on Group page should be ordered with the most recent at the top, going down to the oldest.
Currently it's oldest -> newest.
It should continue to prioritise and order ongoing and upcoming competitions as it seems to already.

image

mauwaz:
Fix #796 

Fix ordering of `ongoing` and `finished` competitions on group page.
- Ongoing competitions that ends first are displayed first
- Finished competitions that ends later are displayed first

**Screenshots**
Screenshot  Finished  Sort

Screenshot  Ongoing  Sort

### DIFF
--- a/app/src/pages/Group/containers/CompetitionsTable.jsx
+++ b/app/src/pages/Group/containers/CompetitionsTable.jsx
@@ -70,24 +70,25 @@ function CompetitionsTable() {
   );
 }
 
+const compareCompetitions = status => (a, b) => {
+  switch (status) {
+    case 'ongoing':
+      return a.endsAt.getTime() - b.endsAt.getTime() || a.startsAt.getTime() - b.startsAt.getTime();
+    case 'finished':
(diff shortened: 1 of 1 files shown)


app/src/pages/Group/containers/CompetitionsTable.jsx
```diff
@@ -74,6 +74,14 @@ function sortCompetitions(competitions) {
   if (!competitions) return [];
 
   return competitions.sort((a, b) => {
+    if (a.status === b.status) {
```
Jonxslays, line 77: This is the same condition being checked on line 86, so if the code passes that comparison it means that the status are guaranteed to be the same because subtracting the indices will result in a falsy 0 value.

I think it would be better to not perform the status comparison twice.
Did you try to fit these changes into the below return block?

Here's something I came up with if you're looking for inspiration:
image

psikoi:
beautiful, merging!

thanks @mauwaz for the contribution, and @Jonxslays for the review

PR #1110, opened 2023-03-24
Add metric leaders to group statistics page

Jonxslays:
This PR adds an additional tab onto the Group Statistics page to view the metric leaders. It also removes the rank column for the average stats as it doesn't actually correspond to any players rank.

image

image


Jonxslays:
I'm wondering if removing the rank column like I did is a good idea. It represents the average rank of the group in each metric, even though it's not any one individuals rank. The table looks pretty sparse without it, though.

psikoi:
> I'm wondering if removing the rank column like I did is a good idea. It represents the average rank of the group in each metric, even though it's not any one individuals rank. The table looks pretty sparse without it, though.

I think the average rank isn't very interesting, so I think it's fine to not have it

PR #1120, opened 2023-04-11
Replace usage of Promise.all for write operations

Jonxslays:
This PR replaces the usage of `Promise.all` in situations where it was being used for write operations.

The following occurrences were not replaced as they are read operations:

- FindPlayerParticipationsStandingsService.ts
- FindGroupDeltasService.ts
- FindPlayerDeltasService.ts
- ImportPlayerHistoryService.ts
- job.manager.ts

### DIFF
--- a/server/src/api/modules/name-changes/services/BulkSubmitNameChangesService.ts
+++ b/server/src/api/modules/name-changes/services/BulkSubmitNameChangesService.ts
@@ -1,3 +1,4 @@
+import { NameChange } from '@prisma/client';
 import { z } from 'zod';
 import { BadRequestError } from '../../../errors';
 import * as playerUtils from '../../players/player.utils';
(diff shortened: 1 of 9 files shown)


server/src/api/modules/name-changes/services/BulkSubmitNameChangesService.ts
```diff
@@ -32,28 +33,24 @@ type BulkSubmitResult = {
 
 async function bulkSubmitNameChanges(payload: BulkSubmitParams): Promise<BulkSubmitResult> {
   const input = inputSchema.parse(payload);
+  const submitted: NameChange[] = [];
```
psikoi, line 36: this could actually just be a number count, since we don't seem to anything with the actual array itself, just it's length

could you include this change please? 🙏

PR #1124, opened 2023-04-13
Update the test script for more fine grained control

Jonxslays:
This PR updates the `run-tests.sh` script to allow for passing a test name and also a test suite to the command.

Examples:

```
# Run all tests
npm run test 

# Run players.test.ts integration and unit test
npm run test players 

# Run only the countries unit test
npm run test countries u

# Run only the snapshots integration test
npm run test snapshots i
```

Note: the `u` and the `i` can be replaced with any of the following and still work:
`U`, `I`, `-u`, `-i`, `-U`, `-I`, `integration`, `unit`

There is currently no support for passing multiple test file names, but that could potentially be added in the future. 
For now just being able to narrow it down to a single file name feels useful enough.

Updated docs for server contribution with these examples.

### DIFF
--- a/.github/contributing/app-guide.md
+++ b/.github/contributing/app-guide.md
@@ -66,7 +66,7 @@ SKIP_PREFLIGHT_CHECK=true
 # REACT_APP_BASE_API_URL=http://localhost:5000/api
 ```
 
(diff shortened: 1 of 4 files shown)


psikoi:
this is beautiful, and thanks for the typo fixes too 😭 

i'll test and merge as soon as I can

PR #1141, opened 2023-05-07
Filter archived players

### LINKED ISSUE #1135: [Bug] Filter out archived players from any global leaderboards
Archived players shouldn't appear in records/deltas/efficiency leaderboards

currently in production, this archived player is showing up:

image

ethanhorrigan:
Closes #1135

### DIFF
--- a/server/__tests__/suites/integration/efficiency.test.ts
+++ b/server/__tests__/suites/integration/efficiency.test.ts
@@ -397,8 +397,6 @@ describe('Efficiency API', () => {
         ehp: 1000
       });
 
-      // Should not contain archived players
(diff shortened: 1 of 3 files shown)


psikoi:
This filter should be done on the SQL query (line 40/41) to make sure the number of results returned by the query is the same as the amount requested by the pagination parameters.

Otherwise you might end up requesting 20 results, but only receiving 18 if 2 of those players were archived.

Also it would be useful to add a integration test case for this in the `efficiency.test.ts` file. (manually add an archived player with high EHP/ehb in `beforeAll`, then when fetching leaderboards for EHP/EHB, make sure no results have the archifed status)

psikoi:
thanks! code looks good! running CI now

PR #1153, opened 2023-05-13
Make player nullable for metric leaders

Jonxslays:
This PR updates metric leaders so that if the leading player after sorting has -1 in the metric/rank the player field is left as `null` indicating no player leads in this metric. 

The front end will display the username as `---` if no player leads.


psikoi:
@Jonxslays could you bump the version to 2.2.31? (has conflicts rn)

PR #1172, opened 2023-06-17
[incomplete] Remove lodash dependency with custom functions

psikoi:
Lodash is bloated and I'm really only using 2 relatively simple functions from it.

The bundle size wasn't as much of a concern when it was only used in the server, but now it also gets bundled into the client-js npm package and that's not good as lodash alone takes up about 55.8% of the package's minified size. (82kb)

I also used to use a lot more lodash functions but have been removing them as I could, but now, 2 functions for 82kb doesn't seem worth it at all.

**This PR is incomplete as it fails to pass tests and I don't want to spend much time on this**, if anyone wants to take a look at what's wrong and push it over the line, that'd be great 🙏


psikoi:
Done in #1190

PR #1181, opened 2023-06-30
[v2 App] Link player and global records to the period they were achieved

### LINKED ISSUE #843: Link records to the gained page for the period the record was obtained
From Majora#8946 on Discord
> Could we see an update to an individual persons' records page so that their records are hyperlinked to the exact 6h/day/week/month they obtained it? For example, a person's overall hiscores, being able to click on the month, with it showing exactly what skills that person trained during that month

Temple does this already, and can be seen in use [here](https://gyazo.com/ba5ba52ff95353b8bfb7bff6771c70ae)

WOM current displays the date that the record was registered, this should link to the individual's profile, showing the gains for the relevant period.

rorro:
Closes #843

### DIFF
--- a/app-v2/src/app/leaderboards/records/page.tsx
+++ b/app-v2/src/app/leaderboards/records/page.tsx
@@ -99,7 +99,11 @@ async function RecordLeaderboard(props: RecordLeaderboardProps) {
                 <PlayerIdentity
                   player={row.player}
                   caption={formatDate(row.updatedAt)}
-                  href={`/players/${row.player.username}/records`}
+                  href={`/players/${row.player.username}/gained/?metric=${
+                    row.metric
+                  }&startDate=${new Date(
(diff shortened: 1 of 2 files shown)


psikoi:
btw we now get preview links for frontend changes (with v2) which is really nice, a few notes:

- I think the whole "box" should become a link, instead of just the "gained value", and have some sort of hover style, maybe just `hover:bg-gray-800` is enough tbh
- you could also add these links to the global record leaderboards, those currently redirect to `/players/:username/records`

rorro:
- Player records: made the whole box clickable
- Global records: Just corrected the link to go to the correct gained time frame. If we want to make the whole box clickable here, we need to make some more changes than just correcting the link because it's a table. If I'm not mistaken.

Also made the link open the correct metric.

PR #1190, opened 2023-07-21
Remove lodash dependency and upgrade prisma

Jonxslays:
This PR removes the server and client-js dependence on lodash, and bumps prisma.

- Removed lodash dependencies in server and client-js.
- Bumped server and client-js versions.
- Bumped prisma to ^4.9.0 (bumps to latest without a harder constraint than ^).
- Removed preview feature from prisma schema due to prisma upgrade.
- Updated `omit` function which was causing test failure by converting `Date`'s to `object`'s.
- Updated `resetDatabase` test function to skip extra prisma property that starts with `$`.
- Updated some tests where new version of prisma is more strict on int values.
- Updated test script to add missing quotations.
- Added `.env.local.example` to app-v2 as it would currently only use the production wiseoldman api url.

The change i made to the `omit` function uses the spread operator rather than `structuredClone`. This is a shallow instead of deep copy, I don't think it will have any side effects but I'd recommend double checking that.

Overall I'd say you should pull this down local and just see if you can find any problems, I tested both app and app v2 running against this new server version and it all seemed fine but a 2nd pair of eyes never hurts.

Supercedes #1172


Jonxslays:
You'll have to clue me in on what this `Type-check Client JS` CI failure indicates.

server/src/api/modules/competitions/services/FetchCompetitionDetailsService.ts
```diff
@@ -1,10 +1,11 @@
-import { omit } from 'lodash';
 import { z } from 'zod';
 import prisma, { modifyPlayer, modifySnapshot } from '../../../../prisma';
+import { omit } from '../../../util/objects';
 import { getMetricValueKey, isComputedMetric, Metric } from '../../../../utils';
 import { NotFoundError } from '../../../errors';
 import { CompetitionDetails } from '../competition.types';
 import * as deltaUtils from '../../deltas/delta.utils';
+import { Snapshot } from '@prisma/client';
```
psikoi, line 8: import this from `../../../../prisma` (as PrismaSnapshot) like modifyPlayer, modifySnapshot, etc

Jonxslays: Autocomplete got me 😆

PR #1199, opened 2023-08-08
Remove axios in client-js

Jonxslays:
This PR removes the dependency on `axios` in the `client-js` project, and uses node's native `fetch` instead.

I've bumped the major version for the project. The main reason for this is that with these changes I've altered the public API removing the ability to use methods like `getRequest` on the `WOMClient` and other the client classes.

This can be changed to continue providing access to those methods, I just figured it is not the goal of the library.

Additional testing would be gladly welcomed. A test framework for this package would be gladly welcomed as well as all tests have to be manually run by hand. That may be a good candidate for a new issue.

**Note**: The experimental warning is issued to the console for each use of `fetch` still.


client-js/src/client.ts
```diff
@@ -7,15 +6,15 @@ import RecordsClient from './clients/RecordsClient';
...
+import HttpClient from './clients/HttpClient';
 
 interface WOMClientOptions {
   apiKey?: string;
   userAgent?: string;
   baseAPIUrl?: string;
 }
 
-export default class WOMClient extends BaseAPIClient {
+export default class WOMClient {
```
psikoi, line 17: WOMClient extended BaseAPIClient so that I could still use the base http client without needing to be tied to an existing "module",

for example, for some undocumented endpoints like `/stats`, I could just do

`new WOMClient().getRequest("/stats")`, or an actual example of this usage:

https://github.com/wise-old-man/discord-bot/blob/35242eb14ce314ffc7ae7aefb5caa5afec1cc63c/src/services/wiseoldman.ts#L160

Jonxslays: That's fair, I'll rework it to continue providing that functionality.

Jonxslays: Ok, all that code in the example file should continue running as is with no changes necessary now.

PR #1241, opened 2023-09-20
Check for p2p xp, not just p2p lvls in isF2p function

autumn-smellegy:
This function was returning true if a player had xp in a p2p skill but didn't have lvl 2, ex 50 thieving xp.


Jonxslays:
good call

PR #1251, opened 2023-10-01
[v2-app] Add Efficiency leaderboard pagination

### LINKED ISSUE #1249: Efficiency Leaderboard is capped to 20
In the beta version, we cannot browse the [Efficiency leaderboard](https://beta.wiseoldman.net/leaderboards/efficiency) beyond the first 20 entries. The previous version had a "load more" feature when you reached the bottom of the page.

Perhaps adding pagination would improve the user experience. Pagination would allow users to skip and browse the leaderboard quicker and more easily. 

Thoughts?

devenini:
This should fix #1249

Ranks are currently hardcoded from the loop index. Not sure if there is a cleaner way.

### DIFF
--- a/app-v2/src/services/wiseoldman.ts
+++ b/app-v2/src/services/wiseoldman.ts
@@ -74,9 +74,12 @@ export const getEfficiencyLeaderboards = cache(
     playerType: PlayerType | undefined,
     playerBuild: PlayerBuild | undefined,
     limit: number,
-    offset: number,
+    offset: number
   ) => {
-    return apiClient.efficiency.getEfficiencyLeaderboards({ metric, country, playerType, playerBuild }, { limit, offset });
+    return apiClient.efficiency.getEfficiencyLeaderboards(
+      { metric, country, playerType, playerBuild },
+      { limit, offset }
+    );
   }
 );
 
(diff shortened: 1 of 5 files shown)


app-v2/src/services/wiseoldman.ts
```diff
@@ -70,11 +70,13 @@ export const getDeltaLeaderboard = cache(
 export const getEfficiencyLeaderboards = cache(
   (
     metric: EfficiencyLeaderboardsFilter["metric"],
+    limit: number,
+    offset: number,
     country?: Country,
     playerType?: PlayerType,
-    playerBuild?: PlayerBuild
+    playerBuild?: PlayerBuild,
```
psikoi, line 77: I see that you made limit and offset come first in the list of args because you can't have optional args before required ones.

But in this case, it'd be best do just change this to be:

```
    metric: EfficiencyLeaderboardsFilter["metric"],
    country: Country | undefined,
    playerType: PlayerType | undefined,
    playerBuild: PlayerBuild | undefined,
    limit: number,
    offset: number,
```

This way there's no optional args, everything else looks good!

devenini: The `country` enum doesn't seem to like the `undefined` value. I have this error in local

```
Invalid enum value for 'country'. You must either supply a valid country code, according to the ISO 3166-1 standard.
```

Any idea?

devenini: Ok should be good

Issue #892, opened 2021-09-22
Feature: Option to hide P2P content on F2P accounts


blakegearin:
Showing P2P skills, bosses, etc. on F2P accounts doesn't seem to provide much value.

psikoi:
@autumn-smellegy interested?

autumn-smellegy:
Yes! @psikoi

devenini:
+1

psikoi:
Implemented in #1263. Closing.

autumn-smellegy:
Skills are now hidden, but p2p bosses and activities still show up on f2p profiles.

PR #1263, opened 2023-10-16
[v2 app] Filter metrics based on player build

devenini:
## Filter metrics based on player build

This PR's purpose is to hide P2P skills on the F2P page (#892) and combat skills on Level 3s builds. The changes are applied to the Overview, Gained, and Achievements page components. These modifications are implemented client-side.

Some potential enhancements could include:

- Directly filtering the metrics from the API.
- Modifying the "Base Stats" achievement to exclude unwanted metrics based on the build type (P2P or combat skills).
- Excluding P2P bosses from F2P boss achievements.

### DIFF
--- a/app-v2/src/components/players/PlayerStatsTable.tsx
+++ b/app-v2/src/components/players/PlayerStatsTable.tsx
@@ -7,8 +7,6 @@ import {
   ActivityValue,
   Boss,
   BossValue,
-  COMBAT_SKILLS,
-  MEMBER_SKILLS,
   Metric,
   MetricProps,
   MetricType,
--- a/app-v2/src/components/players/PlayerStatsTable.tsx
+++ b/app-v2/src/components/players/PlayerStatsTable.tsx
@@ -46,7 +46,7 @@ import {
 } from "../Dropdown";
 
 import TableCogIcon from "~/assets/table_cog.svg";
-import { getBuildContextMetrics } from "~/utils/metrics";
+import { getBuildHiddenMetrics } from "~/utils/metrics";
 
 interface PlayerStatsTableProps {
   player: PlayerDetails;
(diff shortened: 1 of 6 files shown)


app-v2/src/utils/metrics.ts
```diff
@@ -0,0 +1,19 @@
...
+  
+    switch(player.build) {
+        case 'f2p':
+        filteredMetrics = metrics.filter(item => !MEMBER_SKILLS.includes(item.metric));
+        break;
+        case 'f2p_lvl3':
+        filteredMetrics = metrics.filter(item => ![...MEMBER_SKILLS, ...COMBAT_SKILLS].includes(item.metric));
+        break;
+        case 'lvl3':
+        filteredMetrics = metrics.filter(item => !COMBAT_SKILLS.includes(item.metric));
```
psikoi, line 14: instead of assigning to `filteredMetrics` and then returning at the bottom, you can just return the filtered array

```
case 'f2p_lvl3':
  return metrics.filter(item => ![...MEMBER_SKILLS, ...COMBAT_SKILLS].includes(item.metric));
case 'lvl3':
  return metrics.filter(item => !COMBAT_SKILLS.includes(item.metric));
```

app-v2/src/utils/metrics.ts
```diff
@@ -0,0 +1,19 @@
+import { COMBAT_SKILLS, MEMBER_SKILLS, Player } from "@wise-old-man/utils";
+
+export function getBuildContextMetrics(player: Player, metrics: any[]) {
```
psikoi, line 3: Instead of using it like
`const filteredRows = getBuildContextMetrics(player, rows)`, you can use it like:

```
const visibleMetrics = getBuildContextMetrics(player); // where it returns Metric[]
const filteredRows = rows.filter(row => visibleMetrics.includes(row.metric));
```

this should make it a little more flexible to use in the future, and not completely dependent on the rows having a `metric` field inside them.

PS: just noticed, you don't need to pass `player: Player` in the first argument, just `build: PlayerBuild`

app-v2/src/components/players/PlayerStatsTable.tsx
```diff
@@ -7,6 +7,8 @@ import {
   ActivityValue,
   Boss,
   BossValue,
+  COMBAT_SKILLS,
```
psikoi, line 10: i dont think these imports are being used

PR #1295, opened 2023-11-04
Allow configurable base api url in client app

Jonxslays:
The `BASE_API_URL` env var was not being respected when sending requests to the back end.

### DIFF
--- a/app/src/services/wiseoldman.ts
+++ b/app/src/services/wiseoldman.ts
@@ -21,6 +21,7 @@ import { notFound } from "next/navigation";
 export const apiClient = new WOMClient({
   userAgent: "Wise Old Man App (v2)",
   apiKey: process.env.APP_API_KEY,
+  baseAPIUrl: process.env.BASE_API_URL ?? "https://api.wiseoldman.net/v2",
 });
 
 export type TimeRangeFilter = { period: Period } | { startDate: Date; endDate: Date };


Jonxslays:
Many other locations will need to be updated or the overall structure of the WOMClient usage refactored to achieve the goal here. 

Investigating.

Jonxslays:
Ok, this should do it. 

Added a `useWOMClient` hook to use in the client components. 

Should that hook be calling `useState` or should I just return the client directly?

aleccaputo:
> Ok, this should do it.
> 
> Added a `useWOMClient` hook to use in the client components.
> 
> Should that hook be calling `useState` or should I just return the client directly?

Yeah you should just be able to export a const out of the file like a singleton and just import it where you need it. Can remove the hook. 

```
export const womClient = new WOMClient({
  userAgent: "WiseOldMan - App v2 (Client Side)",
  baseAPIUrl: process.env.NEXT_PUBLIC_BASE_API_URL ?? "https://api.wiseoldman.net/v2",
});
```

Jonxslays:
It was just easy to use a hook because it had a defined location already. Wasn't sure where to put a single export for the client.

app/.env.local.example
```diff
@@ -1 +1,2 @@
-# BASE_API_URL="http://localhost:5000"
+BASE_API_URL="http://localhost:5000"
```
psikoi, line 1: do we need both? can't we just have 

`NEXT_PUBLIC_BASE_API_URL="http://localhost:5000"`

?

Jonxslays: yeah we could probably just use that one, you're right.

PR #1336, opened 2023-11-14
Add logic to enable official hiscore link to point to correct leaderboard (#1334)

### LINKED ISSUE #1334: "Open Official Hiscores" link not linking to specific game mode
When selecting the "Open Official Hiscores" dropdown menu option on the player profile, it should link you to that player's game mode, as in, if the player's an ultimate ironman, it should link to the UIM hiscores, etc.

camtblair:
This resolves #1334, just a small change but thought I may as well check it off while looking at the codebase.

It's worth noting that the official hiscore link will also need updating for the league branch as well.

### DIFF
--- a/app/src/app/players/[username]/layout.tsx
+++ b/app/src/app/players/[username]/layout.tsx
@@ -313,21 +313,14 @@ function PlayerAttributes(props: PlayerDetails) {
 }
 
 function getHiscoresURL(displayName: string, playerType: PlayerType) {
-  let url: string;
-
   switch (playerType) {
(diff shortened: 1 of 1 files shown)


psikoi:
> It's worth noting that the official hiscore link will also need updating for the league branch as well.
Good call, would you be down to fix this as well? No need for the switch statement there, it's the same base URL for everyone

camtblair:
Thanks for the feedack, both of those should be taken care of now. I've tried to match the style of the rest of the repository, but let me know if you'd prefer things done differently at all.

PR #1344, opened 2023-11-26
[api] Add image and social link editing for patron groups

psikoi:
- Adds (private endpoint) `/patrons/claim/:discordId`
- Adds the ability to edit `bannerImage`, `profileImage` and `socialLinks` on the `PUT /groups/:id` endpoint
  - Note: Only images hosted on WOM's image server are allowed, so we highly recommend using the website to upload images instead.

@Jonxslays ping ping

### DIFF
--- a/client-js/package-lock.json
+++ b/client-js/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@wise-old-man/utils",
-  "version": "3.1.7",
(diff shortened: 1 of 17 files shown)


Jonxslays:
Ooh cool, I will try to break this while I'm testing and let you know how it goes.

PR #1367, opened 2023-12-07
Fully support CSV endpoints

Jonxslays:
This PR adds public documentation for the group members and competition details CSV endpoints.

Added methods to the typescript client for calling these endpoints as well.


server/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "wise-old-man-server",
-  "version": "2.4.20",
+  "version": "2.4.21",
```
psikoi, line 3: no need to bump versions on the API since this doesn't really require a redeployment

Jonxslays: doesn't it? because i made a new public type and included it in client-js?

Jonxslays:
Ok, the server version bump is reverted.

Issue #1373, opened 2023-12-11
Prevent custom periods before 2013


psikoi:
If you go to a player's "gained" page on the website and select custom period, you can input "1" as the year and submit, it'll pretty much crash the browser.

We should prevent the "custom period" modal from being submitted if the year is < 2013. The same applies for a group's "gained" leaderboards.

PR #1374, opened 2023-12-11
[app] Prevent custom periods before 2013

Sharpienero:
Looking to address #1373 

I wasn't sure if there was already a component that can handle something like this, so I wrote it as a ternary for the time being. If there's already an abstract component that exists that I can use, please point me in that direction and I'll use that instead.

I also wasn't sure if there were any contribution guides for labels and generalized styling choices. I looked around but was able to find anything. I thought this implementation was simple enough in terms of visual queues, but when hitting another error (trying to update too fast), I saw a red X in a toast notification.

### DIFF
--- a/app/src/components/CustomPeriodDialog.tsx
+++ b/app/src/components/CustomPeriodDialog.tsx
@@ -63,35 +63,29 @@ export function CustomPeriodDialog(props: CustomPeriodDialogProps) {
 }
 
 function validateDate(date: DateValue, time: TimeValue, type: string) {
-  console.log("validating date")
   const dateTime = toDate(date, time);
   if (type === typeEnum.START && dateTime.getFullYear() < 2013) {
     setInvalidStartDate(true);
     setStartDate(earliestDate);
-    console.log("1 return false")
     return false;
   }
   if (type === typeEnum.END) {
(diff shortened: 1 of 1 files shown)


Sharpienero:
@psikoi For clarification's sake, by the `same` logic, do you mean making the max range currentYear, or do you mean making the min range 2013? Or both?

Sharpienero:
Added a few conditions: Bad start date, bad end date (e.g in the future) and start date greater than end date. Those will all display an error similar to what you liked before, but now give additional information to the user

Sharpienero:
Is there anything you'd like me to change about this implementation @psikoi?

psikoi:
Sorry for the delay on the review, a bit busy IRL atm.

I do think you have overcomplicated this solution a bit. 

1) The invalid "states" don't need to be stored in state variables, they can be computed on render, this would also remove the need for a "reset", as that computed invalid variable would be recreated anytime a selection happens
2) No need to overcomplicate the labels, you can just hardcode some text in there

If the currently selected start/end date is invalid, show an error label, and prevent the submit button from working (disabled=true)

psikoi:
Closing in favour of https://github.com/wise-old-man/wise-old-man/pull/1507

Issue #1393, opened 2024-01-06
Invalid `+-` sign on name change review context popover


psikoi:
On `ReviewContextTooltip.tsx`, there's a hardcoded `+` label. This shouldn't be added for negative numbers.

An example of this is name change `ID: 325501`  (12421342134 -> Zinti)

![Screenshot 2024-01-06 at 17 17 19](https://github.com/wise-old-man/wise-old-man/assets/3278148/c44d5d87-e820-4435-8aac-00b47c8f97c7)

psikoi:
Fixed by #1398

PR #1398, opened 2024-01-17
Add case for when a ehb / ehp may be negative

arvcork:
Re: https://github.com/wise-old-man/wise-old-man/issues/1393

Changes:

1. Added a case for when EHP may be negative and removed the hardcoded + unless the value is positive.
2. Added a case for when EHB may be negative and removed the hardcoded + unless the value is positive.

Outcome:

1. When a number is added which is negative, there is no + sign suffixed.
2. 
Screenshot 2024-01-17 at 20 40 06

Below is the screenshot of the user which was identified in the issue.


psikoi:
Beautiful, thank you very much

Issue #1388, opened 2024-01-01
[docs] efficiency/leaderboard docs are missing offset and limit query params


rorro:
The `GET /efficiency/leaderboard` endpoint in the documentation is missing offset and limit query params. Should be added in [efficiency-endpoints.mdx](https://github.com/wise-old-man/wise-old-man/blob/75014d9eeed5492aadc58a9da965b2bde633748c/docs/docs/efficiency-api/efficiency-endpoints.mdx).

psikoi:
Fixed by #1399

PR #1399, opened 2024-01-18
Add limit and offset params to efficiency api docs

derektolliver:
Re: https://github.com/wise-old-man/wise-old-man/issues/1388

Tested/confirmed both `limit` and `offset` params were functional for the `/v2/efficiency/leaderboard` endpoint.

Changes:
1. Added `limit` and `offset` parameters to efficiency endpoint docs, following pattern observed on [other endpoints/docs](https://docs.wiseoldman.net/players-api/player-endpoints)


psikoi:
Nice, thank you Derek

PR #1400, opened 2024-01-18
Retain the selected account algorithm when navigation on efficiency page

arvcork:
_Re: #1396_

**Changes:**

1. Introduce a new context which can share the currently selected algorithm when it is selected inside of the `AccountTypeSelector` component.
2. Utilise this context within the layout for (rates).

**Outcome**
When a user now changes the tab on the efficiency page, depending on the account type selected it is carried through and retained when changing.

**Why this approach?**
I probably could of done this in a different approach and I am happy to take feedback and alter this should it need to but I thought, if I do this if there's anywhere else in the application that this kind of behaviour can be found we can utilise this context in order to manage it to pass through the application, so if user sets ironman elsewhere, it'll reflect in here also.

https://github.com/wise-old-man/wise-old-man/assets/146995075/b9904322-b0c3-4a18-a610-0a939880ed10


psikoi:
I don't think this is the correct approach, here's what I suggest:

The links for both tabs are defined on `EfficiencyRatesNavigation.tsx`, but they're hardcoded to just `/ehp` and `/ehb`

What we should do is, on that component, read the current page params (`useParams`), and attach the currently selected type to the link's href

arvcork:
@psikoi thanks for the feedback, yeah completely agree with this and completely slipped my mind when doing it.

Updated to reflect your feedback.

Issue #1392, opened 2024-01-06
Clear "reviewContext" when a name change gets approved


psikoi:
A `reviewContext` field is populated on name changes whenever they get auto-reviewed and skipped (for excessive gains, negative gains, etc).

The same name change can later be approved, but this doesn't clear this `reviewContext` column. And it should.

It seems that on `ApproveNameChangeService.ts (L70)` I did set this to undefined, but it should really be null. A integration test case should also be added to ensure this happens.

derektolliver:
@psikoi Looking to address this one but I have some clarifying questions.

Right now I'm viewing the changes necessary here as:
1. Update L70 on `ApproveNameChangeService.ts` to be set to `null`
2. Implement an integration test in `server/__tests__/suites/integration/names.test.ts` to ensure `approveNameChange` updates `reviewContext` as `null` instead of `undefined`

Besides these, when you say:
> The same name change can later be approved, but this doesn't clear this reviewContext column. And it should.

Is this suggesting some additional change(s)? Such as adjusting other references to `prisma.nameChange` that include `reviewContext` outside of `ApproveNameChangeService`?

psikoi:
> Is this suggesting some additional change(s)? Such as adjusting other references to prisma.nameChange that include reviewContext outside of ApproveNameChangeService?

Nope, the two changes you mentioned should be sufficient

1 - set to null when approved
2 - write a test that creates a name change, populates that column, approves the name change, assert if it's now null

PR #1406, opened 2024-01-20
[api] Set reviewContext to null upon nameChange approval

derektolliver:
Fixes https://github.com/wise-old-man/wise-old-man/issues/1392

Changes:
* Change `approveNameChange` to update `reviewContext` to `null` upon successful resolution
* Add integration test to verify this behavior by creating a name change request, populating the `reviewContext` column, then calling `/names/:id/approve` and verifying `reviewContext` is `null`

Outcome:
* Approved name changes now clear `reviewContext`

**Why this approach?**
I needed to use the `Prism.DbNull` type as opposed to `null` since [Prism appears to have nuance when working with null values](https://www.prisma.io/docs/orm/prisma-client/special-fields-and-types/working-with-json-fields#using-null-values), especially for columns expecting JSON. I think this is probably why `undefined` was originally set here as the error message is not very helpful:

```
Argument reviewContext for data.reviewContext must not be null. Please use undefined instead.
```

### DIFF
--- a/server/src/api/modules/name-changes/services/ApproveNameChangeService.ts
+++ b/server/src/api/modules/name-changes/services/ApproveNameChangeService.ts
@@ -67,7 +67,7 @@ async function approveNameChange(payload: ApproveNameChangeService): Promise<Nam
     data: {
       status: NameChangeStatus.APPROVED,
       resolvedAt: new Date(),
-      reviewContext: undefined
+      reviewContext: PrismaTypes.DbNull
     }
   });
 
--- a/server/__tests__/suites/integration/names.test.ts
+++ b/server/__tests__/suites/integration/names.test.ts
@@ -1086,6 +1086,52 @@ describe('Names API', () => {
         }
       });
     });
+
+    it('should set reviewContext to null upon approval', async () => {
+      const trackResponse = await api.post(`/players/makefrend`);
+      expect(trackResponse.status).toBe(201);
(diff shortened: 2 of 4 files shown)


server/src/api/modules/name-changes/services/ApproveNameChangeService.ts
```diff
@@ -67,7 +67,7 @@ async function approveNameChange(payload: ApproveNameChangeService): Promise<Nam
     data: {
       status: NameChangeStatus.APPROVED,
       resolvedAt: new Date(),
-      reviewContext: undefined
+      reviewContext: PrismaTypes.DbNull
```
psikoi, line 70: Doesn't `null` work? Everything else looks great

derektolliver: Sorry if it wasn't clear in my original phrasing, but when attempting to use `null` that was when I got the error I mentioned:
`Argument reviewContext for data.reviewContext must not be null. Please use undefined instead.`

I put the [documentation/reasoning](https://www.prisma.io/docs/orm/prisma-client/special-fields-and-types/working-with-json-fields#using-null-values) in my original post/description.

psikoi: Oh my bad, I misread the PR's description, sounds good to me

PR #1419, opened 2024-02-06
[app] Redirect all routes to maintenance page in maintenance mode

psikoi:


lezed1:
I normally leave a few tabs of Wise Old Man open for a few different accounts. The redirect caused all those tabs to be directed to https://wiseoldman.net/maintenance! Would it be possible to remove the redirect and maybe return something like a 503 instead so the URL stays intact and any downstream service knows to not cache/crawl Wise Old Man for a bit?

lezed1:
Also, I tried to [join the Discord](https://wiseoldman.net/discord) to see if it had any more information about when maintenance would be finished, but I got redirected back to the maintenance page!

psikoi:
> Also, I tried to [join the Discord](https://wiseoldman.net/discord) to see if it had any more information about when maintenance would be finished, but I got redirected back to the maintenance page!

oof big fail there, sorry, please join us at https://discordapp.com/invite/Ky5vNt2

PR #1452, opened 2024-02-26
[api] Improve TypeScript strictness

psikoi:
- Adds `ts-reset` package (improves typescript defaults)
- Enables `strict: true` on tsconfig for more strict (and hopefully safer 🤞) types
- Adds some required smaller refactors to accommodate the new stricter rules


client-js/package.json
```diff
@@ -18,6 +18,7 @@
   },
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
+  "type": "module",
```
aleccaputo, line 21: Without changing the build to export an ESM compliant bundle, consumers of the package who have also declared type: module will error out as the only module type exported by the package is commonJS

Issue #1459, opened 2024-02-29
JS-Client defining type: module while exporting only CJS


aleccaputo:
![image](https://github.com/wise-old-man/wise-old-man/assets/19272595/0352cfe6-b15f-47ca-8efe-63adaabc84b3)

Noticed this when installing the latest `@wise-old-man/utils`. It's because the `package.json` defines `type: module` which implies the resulting bundled code is ESM compliant.

There are two fixes here:
1. The easy one is to just remove that type declaration in `package.json` (or rename the bundle to `.cjs`)
But ESM is cool and good so....
2. We could have another build step that outputs both a commonJS bundle AND and ESM bundle.

Here's an example of someone doing it adding the required `exports` flag to `package.json`:
https://dev.to/mbarzeev/hybrid-npm-package-through-typescript-compiler-tsc-150c

psikoi:
Is this something you could contribute to @aleccaputo ?

aleccaputo:
@psikoi for sure.
Do you think i should just remove the module declaration for now and make a separate PR for exporting both builds?

psikoi:
> @psikoi for sure. Do you think i should just remove the module declaration for now and make a separate PR for exporting both builds?

yea that sounds like a good plan

psikoi:
Fixed in #1461

PR #1461, opened 2024-03-01
#1459 Build both a cjs and esm bundle

aleccaputo:
Since we are still only building a commonJS bundle and no esm bundle we should remove the module declaration so projects using esm don't blow up.

Future steps:
export both bundles


psikoi:
Can you also bump the package version? I don't think npm allows us to push to the same version

aleccaputo:
yup good call

aleccaputo:
fixes #1461

Issue #1450, opened 2024-02-26
[api] Group recent activity changes


rorro:
We discussed on Discord about showing previous role in the group recent activity. That will require adding a `previousRole` field to [MemberActivity](https://github.com/wise-old-man/wise-old-man/blob/b8ae7d2d4f7333a6c01a0681f608a1ffe3299063/server/prisma/schema.prisma#L586C7-L586C22).

While we are at it, we should also apply the following changes. These would not be shown on the website, but would be nice to have for anyone who is consuming the API.
- For joined group activity, we should populate `role` with the new member's role.
- For left group activity, we should populate `role` with the role of the member that left.

The following types will change.
- [MemberActivity](https://github.com/wise-old-man/wise-old-man/blob/b8ae7d2d4f7333a6c01a0681f608a1ffe3299063/client-js/prisma-models.d.ts#L14): Add previousRole
- [MemberRoleChangeEvent](https://github.com/wise-old-man/wise-old-man/blob/b8ae7d2d4f7333a6c01a0681f608a1ffe3299063/server/src/api/modules/groups/group.types.ts#L83): Only omit createdAt. Remove ` & { previousRole: GroupRole; }`
- [MemberJoinedEvent/MemberLeftEvent](https://github.com/wise-old-man/wise-old-man/blob/b8ae7d2d4f7333a6c01a0681f608a1ffe3299063/server/src/api/modules/groups/group.types.ts#L87-L89): Omit createdAt and previousRole

PR #1470, opened 2024-03-04
#1450 add previous role

aleccaputo:
Add new column for previous role.
Update usages of model.
Also update example.env to support new devs
#1450


server/src/api/modules/groups/services/EditGroupService.ts
```diff
@@ -266,7 +266,7 @@ async function updateMembers(groupId: number, members: Array<{ username: string;
       leftEvents.push(
         ...excessMemberships
           .filter(m => !ignoreFromLeft.includes(m.playerId))
-          .map(m => ({ playerId: m.playerId, groupId: m.groupId, type: ActivityType.LEFT }))
+          .map(m => ({ playerId: m.playerId, groupId: m.groupId, type: ActivityType.LEFT, role: m.role }))
```
aleccaputo, line 269: The integration test is assuming a null here since it used to be omitted from the model. Now that role is required by the models should we expect that role is added to the edit and removed responses as well, or should i force it to be null here?

client-js/prisma-models.d.ts
```diff
@@ -16,6 +16,7 @@ export type MemberActivity = {
   playerId: number;
   type: ActivityType;
   role?: GroupRole;
```
psikoi, line 18: I think technically these should be `GroupRole | null` instead of optional GroupRole 🤔  my bad on that

server/.env.example
```diff
@@ -1,4 +1,10 @@
-CORE_DATABASE_URL=postgresql://postgres:postgres@localhost:5432/wise-old-man_LOCAL?schema=public
```
psikoi, line 1: https://github.com/wise-old-man/wise-old-man/pull/1471

fixed this separately, but thanks for the heads up

server/prisma/schema.prisma
```diff
@@ -584,11 +584,12 @@ model Snapshot {
...
-  playerId  Int
-  type      ActivityType
-  role      GroupRole?
-  createdAt DateTime     @default(now()) @db.Timestamptz(6)
+  groupId      Int
+  playerId     Int
+  type         ActivityType
+  role         GroupRole?
+  createdAt    DateTime     @default(now()) @db.Timestamptz(6)
+  previousRole GroupRole?
```
psikoi, line 592: nitpick but could you move this up 1 line of code? between role and createdAt

psikoi (commented):
The original issues also mention populating `role` for MemberJoined and MemberLeft activities, do you not plan on including these?

client-js/prisma-models.d.ts
aleccaputo: fixed

server/.env.example
aleccaputo: reverted

PR #1481, opened 2024-03-10
Update GitHub Actions

Fdawgs:
This PR bumps GitHub Actions to their latest major versions. Main change is they are now using Node 20 internally instead of 16, which was EOL as of April 2022:
 - Changelog for `actions/checkout` can be [found here](https://github.com/actions/checkout/blob/main/CHANGELOG.md)
 - Releases for `actions/setup-node` can be [found here](https://github.com/actions/setup-node/releases)


psikoi:
Nice! Thank you

Issue #1490, opened 2024-03-22
Mirror in-game role order


psikoi:
If we add a new model to our database that looks something like this:

```typescript
// GroupRoleOrder
{
    groupId: number;
    role: GroupRole;
    index: number;
}
```

We could then populate this information via de "edit group" endpoint (which is used by the RL plugin's "Sync" feature):

`PUT /groups/:id`

```
{
    name: ...
    clanChat: ...
    roleOrder: [
        { role: "magician", index: 0 },
        { role: "legend", index: 1 },
        { role: "owner", index: 2 },
   ]
}
```

and then fetch this information, and return a sorted members list on `GET /groups/:id`

I don't think this would require any frontend work. Unless we decide this should also be configurable in the website, but for now, it seems fine to be a RL Sync feature only.

aleccaputo:
What does the migration for this look like? Is there no migration and only net-new when someone syncs the plugin?

psikoi:
> What does the migration for this look like? Is there no migration and only net-new when someone syncs the plugin?

Yea, I think so

psikoi:
Implemented in #1496 (still needs RL plugin integration)

PR #1496, opened 2024-03-23
#1490 Add Group Role Order (table + api)

aleccaputo:


server/prisma/schema.prisma
```diff
@@ -200,13 +200,30 @@ model Group {
...
 
   // Constraints
   @@unique([name], map: "groups_name_key")
   // Map
   @@map("groups")
 }
 
+model GroupRoleOrder {
+  groupId Int
+  role    GroupRole @unique @default(member)
```
psikoi, line 213: no need for default here, it needs to be a required field

server/prisma/schema.prisma
```diff
@@ -200,13 +200,30 @@ model Group {
   memberActivity MemberActivity[]
   patronages     Patron[]
   socialLinks    GroupSocialLinks[]
+  roleOrder      GroupRoleOrder[]
```
psikoi, line 203: technically should be `roleOrders` since it's an array

server/prisma/schema.prisma
```diff
@@ -200,13 +200,30 @@ model Group {
...
+model GroupRoleOrder {
+  groupId Int
+  role    GroupRole @unique @default(member)
+  index   Int       @unique
+
+  // Relations
+  group Group @relation(fields: [groupId], references: [id], onDelete: Cascade, onUpdate: NoAction)
+
+  // Constraints
+  @@id([groupId])
```
psikoi, line 220: The ID should be a compound key of [groupId, role, index], and the `@unique` tags should be removed from those fields, otherwise there could only be 1 row per groupId, and 1 row per role. See `Membership` as an example.

aleccaputo: yup, sorry hadn't updated this, realized that in my tests

server/prisma/schema.prisma
aleccaputo: done

server/prisma/schema.prisma
aleccaputo: done

aleccaputo:
Should i also add this to the group search response?

psikoi:
> Should i also add this to the group search response (`/groups`)

No, I think this data itself shouldn't be exposed, only used for sorting members. And since group search doesn't expose any members list, it isn't needed

server/src/prisma/index.ts
```diff
@@ -158,6 +159,10 @@ type NameChange = Omit<PrismaNameChange, 'reviewContext'> & {
 
 type GroupSocialLinks = Omit<PrismaGroupSocialLinks, 'id' | 'groupId' | 'createdAt' | 'updatedAt'>;
 
+type GroupRoleOrder = Omit<PrismaGroupRoleOrder, 'index'> & {
```
psikoi, line 162: You don't need this, I only do this to bigint fields (hack for prisma, don't ask 🤦 )

Just export as normal like `Competition`

server/src/api/modules/groups/services/FetchGroupDetailsService.ts
```diff
@@ -29,7 +30,8 @@ async function fetchGroupDetails(id: number): Promise<GroupDetails> {
     // Sort the members list by role
     memberships: group.memberships.sort(
       (a, b) => priorities.indexOf(b.role) - priorities.indexOf(a.role) || a.role.localeCompare(b.role)
-    )
+    ),
+    roleOrders: group.roleOrders
```
psikoi, line 34: Should we expose these? They're most useful just to sort the members list, idk if API consumers need them

server/src/api/modules/groups/services/EditGroupService.ts
```diff
@@ -448,3 +479,20 @@ function calculateRoleChangeMaps(
 }
 
 export { editGroup };
+
+async function updateGroupRoleOrder(
+  groupId: number,
+  roleOrders: Array<{ role: GroupRole; index: number }>,
+  transaction: PrismaTypes.TransactionClient
+) {
+  // TODO: unsure if there's a better way to delete/replace records using prisma
```
Jonxslays, line 488: Would [`upsert`](https://www.prisma.io/docs/orm/reference/prisma-client-reference#upsert) work here?

aleccaputo: thought about it but unsure how it works with prisma, would it look at the compound key to do it since there isn't explicit ids to do it on.

server/__tests__/suites/integration/groups.test.ts
```diff
@@ -1207,6 +1207,54 @@ describe('Group API', () => {
...
+        verificationCode: globalData.testGroupOneLeader.verificationCode,
+        roleOrders: [
+          { role: 'member', index: 1 },
+          { role: 'air', index: 2 }
+        ]
+      });
+
+      expect(response.status).toBe(200);
+
+      const response2 = await api.get(`/groups/${globalData.testGroupOneLeader.id}`).send();
```
psikoi, line 1222: no need for empty `send`.

Also, we should be testing if the order of members is as expected (member -> air)

And we should also test the case that no roleOrders are available, it sorts privileged roles first (see group utils)

server/src/api/modules/groups/services/FetchGroupDetailsService.ts
Jonxslays: Scenario:

I have a discord bot for my clan, and we have a list members command. I want to sort my members in order of role.

I say keep it.

server/__tests__/suites/integration/groups.test.ts
aleccaputo: the order seems like a consumer concern, right? feel like we should just return the set and let the client sort based on index? could be off on that tho

server/src/api/modules/groups/services/FetchGroupDetailsService.ts
psikoi: Sounds good to me

server/src/api/modules/groups/services/FetchGroupDetailsService.ts
```diff
@@ -29,7 +34,8 @@ async function fetchGroupDetails(id: number): Promise<GroupDetails> {
     // Sort the members list by role
     memberships: group.memberships.sort(
       (a, b) => priorities.indexOf(b.role) - priorities.indexOf(a.role) || a.role.localeCompare(b.role)
```
psikoi, line 36: Shouldn't we be sorting the members list here? If roleOrders exist: use those, otherwise fallback to the current behaviour (using `priorities`)

server/src/api/modules/groups/services/EditGroupService.ts
psikoi, line 488: This approach seems fine to me, it's running inside a transaction anyways

server/src/api/modules/groups/services/FetchGroupDetailsService.ts
aleccaputo: updated with new util function

server/__tests__/suites/integration/groups.test.ts
aleccaputo: done

server/src/api/modules/groups/services/EditGroupService.ts
aleccaputo: removed todo

server/prisma/schema.prisma
```diff
@@ -200,13 +200,30 @@ model Group {
...
+
+  // Relations
+  group Group @relation(fields: [groupId], references: [id], onDelete: Cascade, onUpdate: NoAction)
+
+  // Constraints
+  @@id([groupId, role, index])
+  // Indices
+  @@index([groupId], map: "group_role_order_group_id")
+  // Map
+  @@map("groupRoleOrder")
```
psikoi, line 224: can you make change this to plural? to match the other tables

server/src/api/modules/groups/group.types.ts
```diff
@@ -11,6 +11,7 @@ export interface GroupListItem extends Omit<Group, 'verificationHash'> {
 export interface GroupDetails extends GroupListItem {
   socialLinks?: GroupSocialLinks;
   memberships: MembershipWithPlayer[];
+  roleOrders?: GroupRoleOrder[];
```
psikoi, line 14: I wonder if from an API design prespective, it would be better to have this be either an empty array by default, or null by default (instead of non-existant/undefined)

@Jonxslays might have some thoughts

server/src/api/modules/groups/group.utils.ts
```diff
@@ -16,4 +19,20 @@ function buildDefaultSocialLinks() {
...
-export { sanitizeName, buildDefaultSocialLinks };
+function sortMembers(group: Omit<GroupDetails, 'socialLinks' | 'memberCount'>): MembershipWithPlayer[] {
+  if (group.roleOrders && group.roleOrders.length) {
+    // this assumes roleOrders is sorted by index ascending out of the database
+    return [...group.memberships].sort(
+      (a, b) =>
+        (group.roleOrders?.find(order => order.role === a.role)?.index ?? 0) -
+        (group.roleOrders?.find(order => order.role === b.role)?.index ?? 0)
+    );
+  } else {
```
psikoi, line 30: Unnecessary else

server/src/api/modules/groups/group.utils.ts
```diff
@@ -16,4 +19,20 @@ function buildDefaultSocialLinks() {
   };
 }
 
-export { sanitizeName, buildDefaultSocialLinks };
+function sortMembers(group: Omit<GroupDetails, 'socialLinks' | 'memberCount'>): MembershipWithPlayer[] {
+  if (group.roleOrders && group.roleOrders.length) {
+    // this assumes roleOrders is sorted by index ascending out of the database
+    return [...group.memberships].sort(
+      (a, b) =>
+        (group.roleOrders?.find(order => order.role === a.role)?.index ?? 0) -
```
psikoi, line 27: Maybe we could create a map from `group.roleOrders` and then use it for O(1) lookups? Otherwise you'll be iterating that array a lot of times, especially for large groups.

something like

```typescript
const roleOrderMap = new Map(group.roleOrders.map(r => [r.role, r.index]));

return [...group.memberships].sort((a, b) => {
  return (roleOrderMap.get(a.role) ?? 0) - (roleOrderMap.get(b.role) ?? 0) // should these default to 0?
})
```

Also note my code comment above ^ 🤔

server/src/api/modules/groups/services/EditGroupService.ts
```diff
@@ -448,3 +479,19 @@ function calculateRoleChangeMaps(
...
+  transaction: PrismaTypes.TransactionClient
+) {
+  const orderWithGroupId = roleOrders.map(x => ({ ...x, groupId: groupId }));
+
+  await transaction.groupRoleOrder.deleteMany({
+    where: { groupId }
+  });
+
+  await transaction.groupRoleOrder.createMany({
+    data: orderWithGroupId
```
psikoi, line 495: I'd just do `data: roleOrders.map(x => ({ ...x, groupId: groupId }))` and skip creating that variable altogether

server/src/api/modules/groups/group.types.ts
Jonxslays: Either of those would be better than undefined. I would lean towards the empty array.

psikoi: Agreed, and also `socialLinks` shouldn't be undefinedable

aleccaputo: From a non breaking change perspective it seems better to make it optional. Since you don't know like what type of serializes people are using to consume the endpoints? Maybe I'm overthinking it

aleccaputo: done

server/src/api/modules/groups/group.utils.ts
```diff
@@ -16,4 +19,19 @@ function buildDefaultSocialLinks() {
   };
 }
 
-export { sanitizeName, buildDefaultSocialLinks };
+function sortMembers(group: Omit<GroupDetails, 'socialLinks' | 'memberCount'>): MembershipWithPlayer[] {
```
psikoi, line 22: One last nitpick:

Could you change this to `sortMembers(memberships: MembershipWithPlayer[], roleOrders?: GroupRoleOrder[])` instead of receiving the whole group details object and doing an omit type on it?

server/src/api/modules/groups/group.utils.ts
```diff
@@ -16,4 +19,19 @@ function buildDefaultSocialLinks() {
...
+  if (group.roleOrders && group.roleOrders.length) {
+    const roleOrderMap = new Map(group.roleOrders.map(r => [r.role, r.index]));
+    // this assumes roleOrders is sorted by index ascending out of the database
+    return [...group.memberships].sort(
+      (a, b) => (roleOrderMap.get(a.role) ?? 0) - (roleOrderMap.get(b.role) ?? 0)
```
psikoi, line 27: Doesn't defaulting to 0 mean any role that isn't in the order goes to the top?

aleccaputo: good call

server/src/api/modules/groups/group.utils.ts
aleccaputo: done

aleccaputo:
updated csv export

PR #1507, opened 2024-04-30
Prevent submitting custom date range before 2013

nic-w:
Potential solution for #1373 

https://github.com/wise-old-man/wise-old-man/assets/43260193/e8544107-4639-4e89-ba26-f0ea903534aa

### DIFF
--- a/app/src/components/CustomPeriodDialog.tsx
+++ b/app/src/components/CustomPeriodDialog.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import { useEffect, useState } from "react";
+import { useState } from "react";
 import { Time } from "@internationalized/date";
 import { Period, PeriodProps } from "@wise-old-man/utils";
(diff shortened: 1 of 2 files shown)


app/src/components/CustomPeriodDialog.tsx
```diff
@@ -58,6 +70,9 @@ export function CustomPeriodDialog(props: CustomPeriodDialogProps) {
             e.preventDefault();
             handleSelection();
           }}
+          onChange={() => {
```
psikoi, line 73: Could we make these validations on submit instead? Not a huge fan of disabling the confirm button while typing

Maybe by showing a red error message above the button

nic-w: @psikoi updated


https://github.com/wise-old-man/wise-old-man/assets/43260193/511edfd0-f806-490a-8ab1-d75210ee71c7

PR #1526, opened 2024-06-28
Update : Updated Duration dateformat on Competitions Page

godspeed-03:
Feature : Enhanced the duration card to show the timeformat as : 

Additon of year as numeric which is passes as a prop `formatOptions` to `LocalDate` in `CompetitionWidgets` component.

For other cards: As there is a text wrap on the card, the height of all cards had to be increased. Hence, each card is given `h-28` instead of `h-24`


godspeed-03:
Hii @rorro , let check the PR and suggest any changes if nedded.

godspeed-03:
Hii @psikoi, it is important to show the year data in competition widget and also htis PR updates this feature. Have a look, and let's discuss if changes needed.

psikoi:
> Hii @psikoi, it is important to show the year data in competition widget and also htis PR updates this feature. Have a look, and let's discuss if changes needed.

Sorry for the delay on reviewing, i've been very busy.

I'm not a huge fan of the height increase, aesthetically, is there a way to avoid this?

PR #1530, opened 2024-07-09
Add tags for groups

aleccaputo:
Currently using a base set from the OSRS discord.

### DIFF
--- a/client-js/prisma-models.d.ts
+++ b/client-js/prisma-models.d.ts
@@ -70,6 +70,11 @@ export type GroupRoleOrder = {
   index: number;
 };
 
+export type GroupTag = {
+  role: GroupTag;
+  index: number;
+};
+
 type Participation = {
   playerId: number;
   competitionId: number;
--- a/server/prisma/schema.prisma
+++ b/server/prisma/schema.prisma
@@ -201,6 +201,7 @@ model Group {
   patronages     Patron[]
   socialLinks    GroupSocialLinks[]
   roleOrders     GroupRoleOrder[]
+  tags           GroupTag[]
 
   // Constraints
   @@unique([name], map: "groups_name_key")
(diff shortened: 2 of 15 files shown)


server/src/prisma/enum-adapter.ts
```diff
@@ -762,3 +762,26 @@ export const ActivityType = {
 } as const;
 
 export type ActivityType = (typeof ActivityType)[keyof typeof ActivityType];
+
+export const GroupTags = {
```
psikoi, line 766: Isn't this a GroupTagType?

Wouldn't it make more sense for a tag to be (groupId, type, index)?

client-js/prisma-models.d.ts
```diff
@@ -70,6 +70,11 @@ export type GroupRoleOrder = {
   index: number;
 };
 
+export type GroupTag = {
```
psikoi, line 73: Shouldn't this be

```
export type GroupTag = {
  groupId: number;
  type: GroupTagType;
  index: number;
}
```

?

server/prisma/schema.prisma
```diff
@@ -224,6 +225,22 @@ model GroupRoleOrder {
   @@map("groupRoleOrders")
 }
 
+model GroupTag {
+  groupId Int
+  tag     GroupTagType
```
psikoi, line 230: `type` would be more accurate?

server/prisma/schema.prisma
```diff
@@ -224,6 +225,22 @@ model GroupRoleOrder {
...
+  index   Int
+
+  // Relations
+  group Group @relation(fields: [groupId], references: [id], onDelete: Cascade, onUpdate: NoAction)
+
+  // Constraints
+  @@id([groupId, tag, index])
```
psikoi, line 237: actually the constraint should be `[groupId, type]`, since we can't have the same tag type twice for the same groupId.

server/src/api/modules/groups/services/CreateGroupService.ts
```diff
@@ -61,13 +63,23 @@ async function createGroup(payload: CreateGroupPayload): Promise<CreateGroupResu
         createMany: {
           data: memberships
         }
+      },
+      tags: {
+        createMany: {
+          data: payload.tags ?? []
```
psikoi, line 69: Probably needs some validation to make sure there aren't duplicate tags?

server/src/api/modules/groups/services/EditGroupService.ts
```diff
@@ -106,6 +110,19 @@ async function editGroup(groupId: number, payload: EditGroupPayload): Promise<Gr
     }
   }
 
+  if (tags) {
+    const uniqueIndexes = new Set(tags.map(x => x.index));
+    const uniqueTags = new Set(tags.map(x => x.tag));
+
+    if (uniqueIndexes.size < tags.length) {
+      throw new BadRequestError('Role Order must contain unique indexes for each role');
```
psikoi, line 118: message references role orders (the one below as well)

PR #1576, opened 2024-09-10
Add Bryophyta icons

xgerhard:
Hiya, saw these were missing, added the icon from the OSRS highscore page.


psikoi:
I actually made that icon you see on the OSRS hiscores! (Originally for RuneLite) But for some reason I never went back and replaced it with the one I had for WOM.

I suspect there's a few others in the same situation, but I'll merge this one for now, thanks!

PR #1613, opened 2024-11-30
[league] Add achievements for league trophies

rorro:


rorro:
League ended. Maybe next time.

PR #1639, opened 2025-02-10
Feature/openai integration

rodas96:
Add openai external service

Add check offensive name job, cron 5min

### DIFF
--- a/server/package-lock.json
+++ b/server/package-lock.json
@@ -27,6 +27,7 @@
         "express-useragent": "^1.0.15",
         "ioredis": "^5.2.3",
         "js-md5": "^0.8.3",
+        "openai": "^4.83.0",
         "prom-client": "^13.2.0",
         "rate-limiter-flexible": "^2.3.11",
         "winston": "^3.3.3",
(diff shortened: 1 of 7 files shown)


lewislarsen:
I have to admit, I saw this PR, and I wasn't sure what to make of it.

Strictly speaking—is this necessary? We could have a list of inappropriate words and check against them if we really need to. I don't think we need an external dependency or to send information off to OpenAI. The fewer external dependencies, the better, in my opinion.

I appreciate that effort went into putting this PR together, but I think the pool of people typing inappropriate words into fields within WOM is so low that this is a solution looking for a problem.

If I'm missing something here, please let me know!

psikoi:
> I have to admit, I saw this PR, and I wasn't sure what to make of it.
> 
> Strictly speaking—is this necessary? We could have a list of inappropriate words and check against them if we really need to. I don't think we need an external dependency or to send information off to OpenAI. The fewer external dependencies, the better, in my opinion.
> 
> I appreciate that effort went into putting this PR together, but I think the pool of people typing inappropriate words into fields within WOM is so low that this is a solution looking for a problem.
> 
> If I'm missing something here, please let me know!

I actually suggested @rodas96 work on this because it is something me and some mods have discussed in the past.

And you're right, there aren't many people using offensive words on WOM, but unfortunately those few that do really make people upset at WOM and that's not the experience we want.

They typically create groups/competitions with racist or homophobic names, and add whoever they're feuding with to those groups/competitions, these names then show up on people's profiles and/or in-game chat and that's really bad. Some of these users don't really understand how WOM work and could even think we are responsible for this.

Banning abusers is practically impossible due to WOM not being behind a login, so the best we can do is catch abuse early and act quick to discourage it, and for that we need a few different tools such as this one working together to help the moderation team.

As long as we can keep costs down, it should be a useful thing to alert us to any abuse, before the victims of said abuse report it to us.

lewislarsen:
> > I have to admit, I saw this PR, and I wasn't sure what to make of it.
> > Strictly speaking—is this necessary? We could have a list of inappropriate words and check against them if we really need to. I don't think we need an external dependency or to send information off to OpenAI. The fewer external dependencies, the better, in my opinion.
> > I appreciate that effort went into putting this PR together, but I think the pool of people typing inappropriate words into fields within WOM is so low that this is a solution looking for a problem.
> > If I'm missing something here, please let me know!
> 
> I actually suggested @rodas96 work on this because it is something me and some mods have discussed in the past.
> 
> And you're right, there aren't many people using offensive words on WOM, but unfortunately those few that do really make people upset at WOM and that's not the experience we want.
> 
> They typically create groups/competitions with racist or homophobic names, and add whoever they're feuding with to those groups/competitions, these names then show up on people's profiles and/or in-game chat and that's really bad. Some of these users don't really understand how WOM work and could even think we are responsible for this.
> 
> Banning abusers is practically impossible due to WOM not being behind a login, so the best we can do is catch abuse early and act quick to discourage it, and for that we need a few different tools such as this one working together to help the moderation team.
> 
> As long as we can keep costs down, it should be a useful thing to alert us to any abuse, before the victims of said abuse report it to us.

I wasn't aware that it was such a specific problem; that's disappointing to see. People can be utterly cruel to each other behind the thin veil of online anonymity. Given the added information you've provided, I'm all for this feature if it helps curb abuse of WOM.

Thanks for the additional context!

PR #1740, opened 2025-07-25
Add home world to GroupCard

hanselabreu:
This is a small QOL improvement. Was looking for a new clan to join myself and kept having to click the card to see the clan's home world. When choosing a new group the home world is a pretty big determining factor as you want people to be in your same region for improved ping. I followed a similar design pattern for what we have in the group page.

Here's the before:
Screenshot 2025-07-25 191940

Here's the after:
Screenshot 2025-07-25 191917

### DIFF
--- a/app/src/components/groups/GroupCard.tsx
+++ b/app/src/components/groups/GroupCard.tsx
@@ -5,6 +5,8 @@ import { cn } from "~/utils/styling";
 import { Tooltip, TooltipContent, TooltipTrigger } from "../Tooltip";
 
 import VerifiedIcon from "~/assets/verified.svg";
+import GlobeIcon from "~/assets/globe.svg";
+import PeopleIcon from "~/assets/people-2.svg";
 
 export function GroupCard(props: GroupListItem) {
   return (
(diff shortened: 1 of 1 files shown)


app/src/components/groups/GroupCard.tsx
```diff
@@ -34,9 +36,29 @@ export function GroupCard(props: GroupListItem) {
...
-            </span>
+            <div className="mt-1 flex items-center gap-x-3 text-gray-200">
+              <Tooltip>
+                <TooltipTrigger asChild>
+                  <div className="flex items-center">
+                    <PeopleIcon className="mr-1 h-5 w-5 text-gray-200" />
+                    <span className="text-xs text-gray-200">{props.memberCount}</span>
+                  </div>
+                </TooltipTrigger>
+                <TooltipContent>{props.memberCount} members</TooltipContent>
```
Jakesterwars, line 47: Since it was this way before, swap the members text to do `<TooltipContent>{props.memberCount} members {props.memberCount === 1 ? "member" : "members"}</TooltipContent>` like above or just slap a `member(s)` text. Overall design looked good.

PR #1827, opened 2025-11-29
Add Sailing EHP rates

Paatus:
Adds sailing EHP rates to the API.

Rates are taken from https://templeosrs.com/rates/skills.php


Paatus:
> Could you move the sailing EHP down to be at the bottom instead of at the top? The PR is good to be merged otherwise.

Absolutely, Done!

PR #1936, opened 2026-06-08
Feature/delta in linechart

### LINKED ISSUE #1846: Cumulative Experience Gained - On snapshot hover, show delta from first snapshot
Image

The hovered snapshot could show the delta since first snapshot in the period.

Orinite:
fixes https://github.com/wise-old-man/wise-old-man/issues/1846

https://github.com/user-attachments/assets/4f8c3b93-782d-42b5-ad37-280e08609c78


psikoi:
I don't think it should be showing deltas by default - @Aluminoti is that what you meant?

Aluminoti:
> I don't think it should be showing deltas by default - @Aluminoti is that what you meant?

Yeah, next question to ask related to the info is "but how much since the start of the period that I'm viewing exactly?"

Dedicated toggle or something would be fine but I would argue swapping between the two views makes it tedius to answer the immediate question, or needing to find a spot to enable it to show on the graph could be ugly.

Next next question is Delta since previous point so might even be worth adding that as well

psikoi:
I don't think a toggle is the right move here - maybe the tooltip could show both?
(excuse the shitty rushed design)

image

PR #1938, opened 2026-06-08
Add the Joined At Date of clan Member to group sync 

Orinite:


server/prisma/schema.prisma
```diff
@@ -180,11 +180,12 @@ model GroupRoleOrder {
...
-  groupId   Int
-  role      GroupRole @default(member)
-  createdAt DateTime  @default(now()) @db.Timestamptz(6)
-  updatedAt DateTime  @updatedAt @db.Timestamptz(6)
+  playerId            Int
+  groupId             Int
+  role                GroupRole  @default(member)
+  createdAt           DateTime   @default(now()) @db.Timestamptz(6)
+  updatedAt           DateTime   @updatedAt @db.Timestamptz(6)
+  clientSyncJoinedAt  DateTime?  @db.Timestamptz(6)
```
psikoi: nit: can you move this in between role and createdAt?

server/src/api/modules/groups/services/EditGroupService.ts
```diff
@@ -35,7 +35,7 @@ interface EditGroupPayload {
     twitch?: string | null;
     youtube?: string | null;
   };
-  members?: Array<{ username: string; role: GroupRole }>;
+  members?: Array<{ username: string; role: GroupRole; clientSyncJoinedAt?: string | null }>;
```
psikoi: why nullable AND optional? i'd say we should get rid of the nullable

psikoi: Also, I'm seeing this type repeated a bunch, maybe we should add a type for it in `server/src/types/*`? Maybe something like `GroupMemberInput`, and then re-use it on `AddMembersService` , `CreateGroupService` ,`EditGroupService` ?

server/src/api/responses/membership.response.ts
```diff
@@ -11,5 +11,5 @@ import { pick } from '../../utils/pick.util';
 export type MembershipResponse = Membership;
 
 export function formatMembershipResponse(membership: Membership): MembershipResponse {
-  return pick(membership, 'playerId', 'groupId', 'role', 'createdAt', 'updatedAt');
+  return pick(membership, 'playerId', 'groupId', 'role', 'createdAt', 'updatedAt', 'clientSyncJoinedAt');
```
psikoi: nit: can you move this to in between role and createdAt please?

server/src/api/modules/groups/services/EditGroupService.ts
```diff
@@ -469,15 +494,20 @@ async function addMissingMemberships(
...
   missingPlayers: Player[],
-  memberInputs: Array<{ username: string; role: GroupRole }>
+  memberInputs: Array<{ username: string; role: GroupRole; clientSyncJoinedAt?: string | null }>
 ) {
   const roleMap: { [playerId: number]: GroupRole } = {};
+  const clientSyncJoinedAtMap: { [playerId: number]: Date | null } = {};
 
   missingPlayers.forEach(player => {
     const role = memberInputs.find(m => standardizeUsername(m.username) === player.username)?.role;
+    const clientSyncJoinedAt = memberInputs.find(
```
psikoi: Can probably refactor this to
```typescript
const matchingInput = memberInputs.find(m => standardizeUsername(m.username) === player.username);

if(!matchingInput) {
    continue;
}

const { role, clientSyncJoinedAt } = matchingInput
```

just to avoid iterating that twice

server/__tests__/suites/integration/groups.test.ts
```diff
@@ -138,7 +138,7 @@ describe('Group API', () => {
 
       expect(response.status).toBe(400);
       expect(response.body.message).toMatch(
-        'Invalid members list. Must be an array of { username: string; role?: string; }.'
+        'Invalid members list. Must be an array of { username: string; role?: string; clientSyncJoinedAt?: number | null; }.'
```
psikoi: I don't think we should "promote" that this field exists for **writes** - ideally we'd discourage writing to it outside of client syncs

server/src/api/modules/groups/services/EditGroupService.ts
```diff
@@ -381,6 +384,28 @@ async function updateMembers(groupId: number, members: Array<{ username: string;
 
       const roleUpdatesMap = calculateRoleChangeMaps(keptPlayers, memberships, members);
 
+      // Update clientSyncJoinedAt if needed for existing memberships that don't have a clientSyncJoinedAt date yet.
+      for (const mem of members) {
```
psikoi: I think there's probably a faster implementation out there that uses maps instead of two nested iterations - take a look at `calculateRoleChangeMaps`

server/src/api/modules/groups/services/EditGroupService.ts
Orinite: The plugin sends it null if it isnt in the context of the sync button. Yes I can make a type for this.

server/src/api/responses/membership.response.ts
Orinite: Done

server/src/api/modules/groups/services/EditGroupService.ts
Orinite: Made the type

server/prisma/schema.prisma
Orinite: Done

server/src/api/modules/groups/services/EditGroupService.ts
Orinite: Yes, done!

server/__tests__/suites/integration/groups.test.ts
Orinite: Fair enough, removed from the message

server/src/api/modules/groups/services/EditGroupService.ts
Orinite: Updated to do similar to what that roleChangeMaps function does.

server/src/api/modules/groups/services/EditGroupService.ts
Orinite: I actually tested this in my local changes to plugin. It looks like the Gson configuration does not serialize fields that are null, so its fine to remove the null option from this.

server/src/api/util/validation.ts
```diff
@@ -121,7 +121,8 @@ const groupRoleSchema = z.nativeEnum(GroupRole);
 export const memberSchema = z.object(
   {
     username: z.string(),
-    role: z.optional(groupRoleSchema).default(GroupRole.MEMBER) as unknown as typeof groupRoleSchema
+    role: z.optional(groupRoleSchema).default(GroupRole.MEMBER) as unknown as typeof groupRoleSchema,
+    clientSyncJoinedAt: z.optional(z.string())
```
psikoi: this doesn't validate that it's a ISO date string, should use `getDateSchema` instead

server/__tests__/suites/integration/groups.test.ts
```diff
@@ -932,6 +932,29 @@ describe('Group API', () => {
       });
     });
 
+    it('should allow clientSyncJoinedAt to be set', async () => {
+      const before = await api.get(`/groups/${globalData.testGroupNoLeaders.id}`);
```
psikoi: in here you're fetching `testGroupNoLeaders`, later you're fetching `testGroupOneLeader` - it should be the same in both fetches

server/src/api/modules/groups/services/EditGroupService.ts
```diff
@@ -544,3 +566,30 @@ function calculateRoleChangeMaps(
 
   return newRoleMap;
 }
+
+function calculateClientSyncJoinedAtMaps(
+  keptPlayers: Player[],
```
psikoi: keptPlayers seems unused?

server/src/api/util/validation.ts
Orinite: Makes sense to me, fixed that.

server/__tests__/suites/integration/groups.test.ts
Orinite: oops, good catch, fixed

server/src/api/modules/groups/services/EditGroupService.ts
Orinite: removed

server/src/types/group-member-input.type.ts
```diff
@@ -0,0 +1,7 @@
+import { GroupRole } from './group-role.enum';
+
+export type GroupMemberInput = {
+  username: string;
+  role: GroupRole;
+  clientSyncJoinedAt?: Date | null;
```
psikoi: why null? doesn't the code you wrote in EditGroupService mean a sync date cannot be unset?

imo, we either:
- allow passing in "null" to unset a date
- dont allow nulls, and keep it immutable once set

i think the first one is more flexible, but if you prefer the second i wouldn't oppose

Orinite: The current logic is assuming a value is present and null is not passed in. I think that once it is set, it doesn't make sense to "clear" it from a usability perspective. People aren't leaving the clan and re-joining within a short timeframe that they want it to reflect in the join date in WOM I would think? Maybe I am wrong though on that.

For now, I have done the second option. If we find that we want the flexibility then I am happy to do the first option.

PR #1962, opened 2026-06-29
feat(groups): adding Group Competition Code

antzangell:
## Rationale
Clan leaders currently have to share their main group verification code with anyone who needs to create competitions. This gives those members full admin access. This means they can edit, delete, or modify the group. There's no way to delegate competition creation without giving away the keys to everything, Which personally i feel is too risky as you want clan founders / admin to manage the group and then you can have event organisers / others to have the group competition code allowing to do as such without implicating the group control itself. 

 ## Feature
A secondary "competition code" that can only be used to create group competitions.
Group owners generate/reset/delete it via the group edit page or API, using their main verification code.

New endpoints:
- PUT /groups/:id/competition-code (generate/reset)
- DELETE /groups/:id/competition-code (revoke)

## Desired Outcomes
- Event managers can create group competitions without full group admin access
- Group owners retain full control, they can reset or revoke the competition code at any time
- The main verification code continues to work for everything, including creating competitions (Both codes can be used to make a competition to ensure less annoyance on the consumers end)
- Clear separation of concerns: competition code cannot edit, delete, or manage the group

### DIFF
--- a/client-js/package.json
+++ b/client-js/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@wise-old-man/utils",
-  "version": "4.0.16",
+  "version": "4.1.0",
   "description": "A JavaScript/TypeScript client that interfaces and consumes the Wise Old Man API, an API that tracks and measures players' progress in Old School Runescape.",
   "keywords": [
     "wiseoldman",
--- a/client-js/src/clients/GroupsClient.ts
+++ b/client-js/src/clients/GroupsClient.ts
@@ -220,4 +220,28 @@ export default class GroupsClient extends BaseAPIClient {
   getMembersCSV(id: number) {
     return this.getText(`/groups/${id}/csv`);
   }
+
+  /**
+   * Generates or resets the competition code for a group.
+   * This code allows creating group competitions without full group admin access.
+   * Requires the main group verification code.
+   * @returns The newly generated competition code.
+   */
+  generateCompetitionCode(id: number, verificationCode: string) {
+    return this.putRequest<{ competitionCode: string }>(`/groups/${id}/competition-code`, {
+      verificationCode
+    });
+  }
+
+  /**
+   * Deletes the competition code for a group.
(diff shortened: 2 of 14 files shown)


app/src/components/groups/CompetitionCodeSection.tsx
```diff
@@ -0,0 +1,257 @@
...
+import { Alert, AlertDescription } from "../Alert";
+import { Button } from "../Button";
+import { Dialog, DialogContent, DialogHeader, DialogTitle } from "../Dialog";
+import { Input } from "../Input";
+
+import ClipboardIcon from "~/assets/clipboard.svg";
+import LoadingIcon from "~/assets/loading.svg";
+import WarningIcon from "~/assets/warning.svg";
+
+const MIN_WAIT_PERIOD_SECONDS = 10;
```
antzangell: Is this too annoying? or do we want to add a dismiss option?

antzangell:
image
image
image
image
UI Pictures.

psikoi:
I really appreciate the effort you've put into this, but I don't think adding a new verification code is the right solution. Unfortunately the verification code system is a bad system that I regret and this would just be a band-aid on top that would take effort to maintain and one day migrate off of.

We do have plans on how to handle different levels of permissions, here's a basic outline:
- We introduce Discord oauth
- We invite group admins to link their groups with their Discord account
  - This invalidates the verification code, it can no longer be used
- Group permissions are then able to be granted to others (admin can invite me and grant me "competition.create" and "group.add_member" but not "group.delete")

Doing it through user accounts solves a lot of problems for us, such as:
- people losing their vcode, then asking us to reset (moderation workload)
- people giving their vcode to the wrong person, then asking us to reset (moderation workload)
- groups/comps are created by annonymous users at the moment, which means they can be abused for harassment and we don't have much control over how to prevent future instances, because there is no user to ban
- once there's a link between discord user and wom group, our Discord bot can become a lot more useful and we can add commands such as "add-member"

PR #1976, opened 2026-07-04
[server/client-js/docs] Add bulk group hiscores endpoint

psikoi:


SaltisRS:
oh right as i was about to deploy the other bulk endpoint change to my codebase, suppose ill hold off <3

Issue #1791, opened 2025-09-18
Filtering leaderboards for "F2P" errors out


grubling:
When filtering the playerBuild for f2p or f2p_lvl3, the leaderboard errors out with the following screen:

<img width="1010" height="380" alt="Image" src="https://github.com/user-attachments/assets/39dd3ac2-5dfe-437e-a9ce-901ddc993f6b" />

Link: https://wiseoldman.net/leaderboards/top?playerBuild=f2p

Reproduced on both iphone and windows. Non-f2p filters seem to work fine.

psikoi:
> This seems to be working again. Possibly some changes recently with the hiscores fixed it alongside other updates for WOM.

true, although it's still quite slow and it wouldn't surprise me if it timed out during peak hours. It probably needs some better querying - if there's anyone out there with some SQL skills that would be willing to give it a try 🤷‍♂️

kirypto:
I’ve been digging into the slow F2P leaderboard filtering on this issue and have a spike that improves performance dramatically, but with a denormalization tradeoff. I’d like thoughts from maintainers here before I spend more time finalizing it and opening a PR.

I replicated the issue by seeding a bunch of random data into my local database and testing the endpoint. Once I had it working acceptably, I reset back to master, ran it there, then applied my fix and ran again for comparison. Here's the comparison of `/leaderboards/top?playerBuild=f2p`:

<img width="1920" height="992" alt="Image" src="https://github.com/user-attachments/assets/7cd69895-db61-4183-8e91-3a57956b6241" />

| Stage | Approx. response time |
|--------|------------------------|
| Empty database | ~100–130 ms |
| After seeding ~1M players / ~3M `cachedDeltas` rows | ~6.5 s |
| After denormalizing player fields onto `cachedDeltas` | ~170–260 ms |

On the live API I also hit `GET /v2/deltas/leaderboard?metric=overall&period=day&playerBuild=f2p` and it took on the order of **~10s**, so this is comparable, but if you give me a count of players I could generate more to match.

### Root Cause

The slow path is roughly: find top `cachedDeltas` by `(period, metric, value DESC)`, **then** filter via a join to `players` (`build`, `status`, and optionally `type` / `country`).

[...]

PR #1979, opened 2026-07-07
fix(leaderboards): handle API errors in top leaderboard page gracefully

yakubka:
When a user selects the F2P (free-to-play) player build filter on the top leaderboards page, the underlying API call fails because that combination of filters returns an error response rather than an empty list. Since TopLeaderboard is an async server component and the error propagates uncaught, the entire page crashes with an unhandled exception.

This was reported in issue #1791.

The fix wraps each getDeltaLeaderboard call in a try-catch block, storing a fetchError flag when the call fails. Each panel then renders a graceful inline error message instead of crashing the page. The other panels, which may succeed with the same filters, continue to render normally.

A new error.tsx file was added as a Next.js App Router error boundary for the route segment. This catches any rendering errors that slip past the per-panel try-catch and shows a user-friendly fallback rather than the default error page.

Changes:

app/src/app/leaderboards/top/page.tsx: wrapped getDeltaLeaderboard call with try-catch, added fetchError state, added conditional rendering for the error case alongside the existing empty-state check.

app/src/app/leaderboards/top/error.tsx: new file, use client error boundary for the leaderboards top route segment.

Yokubjon Nurullaev identified the root cause by tracing the unhandled exception from the F2P filter through the async server component to the missing error boundary. He implemented the per-panel try-catch pattern in page.tsx, the fetchError state, and the conditional rendering logic.

IbrokhimN designed the error.tsx boundary component and reviewed the error state logic to ensure the three period panels remain independent, so a failure in one panel does not suppress results from the others.

### DIFF
--- a/app/src/app/leaderboards/top/error.tsx
+++ b/app/src/app/leaderboards/top/error.tsx
@@ -1,0 +1,9 @@
+"use client";
+
+export default function TopLeaderboardsError() {
+  return (
+    <div className="w-full rounded border border-gray-700 py-10 text-center text-sm text-gray-200">
+      Something went wrong while loading the leaderboards. Try adjusting your filters.
+    </div>
+  );
+}
--- a/app/src/app/leaderboards/top/page.tsx
+++ b/app/src/app/leaderboards/top/page.tsx
@@ -72,18 +72,29 @@ interface TopLeaderboardProps {
 async function TopLeaderboard(props: TopLeaderboardProps) {
   const { period, filters } = props;
 
-  const data = await getDeltaLeaderboard(
-    filters.metric,
-    period,
-    filters.country,
-    filters.playerType,
-    filters.playerBuild,
-  );
+  let data: Awaited<ReturnType<typeof getDeltaLeaderboard>> = [];
+  let fetchError = false;
+
+  try {
+    data = await getDeltaLeaderboard(
+      filters.metric,
+      period,
+      filters.country,
+      filters.playerType,
+      filters.playerBuild,
+    );
+  } catch {
+    fetchError = true;
+  }
 
   return (
     <div>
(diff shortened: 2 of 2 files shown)


Jakesterwars (commented):
This may be more of a personal preference for readability, but I would wrap the try catch around the const data portion instead of setting up `let` to store information and check below. You could then have the catch render the errored state still while keeping the call uniform.

Another option is just to `.then` the awaited call for data and then do something like this:

```
const result = await getDeltaLeaderboard(filters.metric,
      period,
      filters.country,
      filters.playerType,
      filters.playerBuild).then(
  (data) => ({ data, error: false }),
  () => ({ data: [], error: true }),
);

const { data, error } = result;
```

and then you can setup the rendering code to below:

```
return (
    <div>
      <h3 className="pb-3 text-h3 font-bold">{PeriodProps[period].name}</h3>
      {error ? (
        <div className="w-full rounded border border-gray-700 py-10 text-center text-sm text-gray-200">
          Could not load leaderboard data for this filter.
        </div>
      ) : data.length === 0 ? (
        <div className="w-full rounded border border-gray-700 py-10 text-center text-sm text-gray-200">
          No results were found
        </div>
      ) : (
        <ListTable>
          {data.map((row, index) => (
            <ListTableRow key={row.player.username}>
              <ListTableCell className="w-1 pr-1">{index + 1}</ListTableCell>
              <ListTableCell>
                <PlayerIdentity
                  player={row.player}
                  href={`/players/${row.player.username}/gained?metric=${filters.metric}&period=${period}`}
                />
              </ListTableCell>
              <ListTableCell className="w-5 text-right font-medium">
                <FormattedNumber value={row.gained} colored />
              </ListTableCell>
            </ListTableRow>
          ))}
        </ListTable>
      )}
    </div>
  );
```

Again, could be personal on that because this works fine the way you have it regardless.

PR #2007, opened 2026-08-14
Add CLAUDE.md contributor guide

Jonxslays:
## Summary

Adds a `CLAUDE.md` at the repo root documenting the codebase for contributors (and AI coding assistants like Claude Code).

It covers:
- **Repository layout** — the four components (`server/`, `app/`, `client-js/`, `docs/`) and the fact that there's no root workspace tooling.
- **Commands** — per-component `dev`/`build`/`lint`/`test` scripts, including the server's `test <name> u|i` shortcuts.
- **Server architecture** — the Router → Service → Prisma flow, module structure, background jobs, and Prisma/types conventions.
- **App architecture** — App Router layout, API service boundary, and component conventions.
- **Conventions** — Prettier/ESLint enforcement, file-naming rules, and a Windows CRLF gotcha.

No code changes — documentation only. Opening this separately so it can be evaluated on its own; feel free to close if it's not something you'd like to maintain in-tree.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,90 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## Repository layout
+
+Monorepo for the Wise Old Man — an Old School RuneScape progress tracker. Each component has its own
+`package.json`, `eslint.config.mjs`, and independent build/deploy pipeline. There is **no root workspace
+tooling** (root `package.json` only holds `prettier`). Always `cd` into the component you're working on.
+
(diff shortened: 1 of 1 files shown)


CLAUDE.md
```diff
@@ -0,0 +1,90 @@
+# CLAUDE.md
```
psikoi, line 1: In general I think this is way too verbose and most of it can be discovered by Claude on its own.

I'd prefer if this focuses on weird or very specific WOM things rather than say things like "`npm run lint` runs the linter"

Jonxslays: it definitely can, but i think the idea is that claude can read this file *without* having to go search the repo and get instant context. it burns less tokens.

psikoi: yes, but you're also injecting all the tokens in this file for every new session - and not every session needs to know about zod route validators (example) - there's a trade off

PR #2008, opened 2026-08-22
[app] Move group and competition deletion into a Danger zone on the edit pages

### LINKED ISSUE #1837: Move the button to delete a group into the edit page
Something like this should help prevent people from accidentally deleting their groups quickly.

Additionally the text could be made red and a few layers of confirmation could be added when trying to delete a group (like if you've ever tried to prestige in MW2).

Jonxslays:
Closes #1837.

## Summary

Moves the "delete" action for both **groups** and **competitions** out of the details actions dropdown (where it was easy to trigger accidentally) and into a new **Danger zone** section on the respective **edit** page, behind the existing verification-code gate.

### Groups

- Removed the **Delete** item from the group details `...` actions menu.
- Added a **Danger zone** entry to the edit page side navigation, styled in red (`text-red-500`, matching the active accent bar).
- The Danger zone is presented as a card (reusing the existing `Alert` `error` variant) with the deletion warning and a red **Delete group** button.
- The delete flow itself is unchanged — the existing `DeleteGroupDialog` still requires re-entering the verification code before anything is deleted.

### Competitions

- Removed the **Delete** item from the competition details `...` actions menu.
- Added a matching **Danger zone** entry to the competition edit page side navigation (both the desktop list and the mobile tabs), styled in red.
- The Danger zone reuses the same `Alert` `error` card with a red **Delete competition** button.
- The delete flow is unchanged — the existing `DeleteCompetitionDialog` still requires re-entering the verification code before anything is deleted.

Because both edit pages are already gated by a verification-code check, the delete entry points are only reachable after owner verification.

## Screenshots

**Group edit page – General section (note the red "Danger zone" nav item):**

image

**Group edit page – Danger zone section:**

image

**Group details actions menu – "Delete" removed (only Edit / Create group competition remain):**

image

### DIFF
--- a/app/src/app/competitions/[id]/(details)/layout.tsx
+++ b/app/src/app/competitions/[id]/(details)/layout.tsx
@@ -11,7 +11,6 @@ import { Alert, AlertDescription, AlertTitle } from "~/components/Alert";
 import { Button } from "~/components/Button";
 import { CompetitionDetailsNavigation } from "~/components/competitions/CompetitionDetailsNavigation";
 import { CompetitionPreviewWarning } from "~/components/competitions/CompetitionPreviewWarning";
-import { DeleteCompetitionDialog } from "~/components/competitions/DeleteCompetitionDialog";
 import { ExportCompetitionDialog } from "~/components/competitions/ExportCompetitionDialog";
 import { PreviewMetricDialog } from "~/components/competitions/PreviewMetricDialog";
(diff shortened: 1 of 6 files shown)


psikoi:
Maybe we could take it one step further and require people to type the group name in some input? (like github does)

Jonxslays:
> Maybe we could take it one step further and require people to type the group name in some input? (like github does)

Done, blocks the delete until they type the name/title and verification code now.

image

image

PR #2021, opened 2026-09-06
Updated EHP rates to match Hallowed Sepulchre changes

Meantub:
Hallowed Sepulchre changed their requirements for floor 4 and 5. I'd assume that would lower the requirements for the EHP calculations. If I'm wrong or am using the wrong XP lemme know.

Floor 4 reqs changed to level 77
Floor 5 reqs changed to level 87


rorro:
Wouldn't we want to set it to level 72 and 82 since the levels required to enter the floors are now boostable as well?

